PR #35, opened 2024-10-13
feat(deposit): implement deposit machine and UI components

jobotics:
- Add UI deposit machine inheriting core concepts
- Refactor deposit machine to work in conjunction with UI machine
- Update components of deposit widget and form

Additional changes:
- Add base component for select


src/components/Select/styles.css
```diff
@@ -0,0 +1,115 @@
+@import "@radix-ui/colors/black-alpha.css";
+@import "@radix-ui/colors/mauve.css";
+@import "@radix-ui/colors/violet.css";
+
+/* reset */
+button {
+	all: unset;
+}
```
cawabunga-bytes, line 8: It is not suggested overriding global styles in component style.

src/components/Select/styles.css
```diff
@@ -0,0 +1,115 @@
+@import "@radix-ui/colors/black-alpha.css";
+@import "@radix-ui/colors/mauve.css";
+@import "@radix-ui/colors/violet.css";
```
cawabunga-bytes, line 3: I have concerns about these styles, not sure if it is necessary to import these styles on component level. For example, another may also want to import them, so we'll end up with duplicates. Additionally these styles change global. I think it is better to move them to sdk global style.

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import {
+  CheckIcon,
+  ChevronDownIcon,
+  ChevronUpIcon,
+} from "@radix-ui/react-icons"
+import * as RadixSelect from "@radix-ui/react-select"
+import clsx from "clsx"
+import type React from "react"
+import "./styles.css"
```
cawabunga-bytes, line 11: Let's use css modules as it helps to isolate class names.

src/features/deposit/components/DepositFormNetwork.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const {
+    handleSubmit,
+    register,
+    setValue,
+    watch,
+    formState: { errors },
+  } = useForm<DepositFormNetworkValues>({ reValidateMode: "onSubmit" })
+
+  return (
+    <div className="w-full md:max-w-[472px] rounded-[1rem] p-5 shadow-paper bg-white dark:shadow-paper-dark dark:bg-black-800">
```
cawabunga-bytes, line 25: We're mixing Tailwind with plain css again. Maybe let's stick to one so it will be easier to migrate in the future?

src/features/machines/depositUIMachine.ts
```diff
@@ -0,0 +1,123 @@
...
+              type: "clearError",
+            },
+          ],
+        },
+      },
+      states: {
+        idle: {},
+      },
+    },
+    genereting: {
```
cawabunga-bytes, line 81: nit: typo genereting -> generating

src/features/deposit/components/DepositFormNetwork.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Form<DepositFormNetworkValues>
+        handleSubmit={handleSubmit((values: DepositFormNetworkValues) =>
+          onSubmit(values)
+        )}
+        register={register}
+      >
+        <div className="mb-4">
+          <Select<BlockchainEnum>
+            options={blockchains}
+            value={watch("blockchain")}
```
cawabunga-bytes, line 35: I think we should use here `register` from react-hook-form, so we won't need manual control over the field.

src/features/machines/depositMachine.ts
```diff
@@ -1,178 +1,127 @@
...
+  context: ({ input }: { input: Input }) => {
+    return {
+      depositRef: null,
+    }
   },
-  id: "deposit_machine",
-  initial: "Idle",
+  id: "deposit",
+  type: "parallel",
```
cawabunga-bytes, line 51: If I understand correctly, these aren't parallel states. You either follow the Near path or the Passive deposit path (depositing to the generated address).

src/features/machines/depositUIMachine.ts
cawabunga-bytes, line 1: I suppose it is still in wip state, right?

cawabunga-bytes (commented):
I left some comments.

src/components/Select/Select.tsx
jobotics: We are considering customizing the theme of our widget so in this case CSS Module wouldn't allow us to use dynamic config. I assume that we have to use CSS-in-JS which is a more robust choice for the kind of customization.

src/features/machines/depositUIMachine.ts
jobotics: Yeap, correct

src/features/machines/depositMachine.ts
jobotics: I see several reasons to do this in different threads:
1. These states have different inputs and outputs
2. When adding X path (Ethereum or other) we won't have to make changes to the current threads

cawabunga-bytes (approved):

jobotics:
FYI: The Deposit machine will be split into `depositViaNearMachine` and `generateDepositAddressMachine`.

PR #36, opened 2024-10-15
feat(deposit): integrate deposit machine

jobotics:
This commit introduces the DepositWidget component and associated functionality, utilizing a state machine for managing the deposit flow. Key changes include:

- Refactor DepositWidget component with form type selection
- Implement DepositFormController for managing different deposit form types
- Create DepositFormNear component for NEAR deposits
- Introduce depositNearMachine for managing deposit state and actions
- Remove depositMachine
- Add DepositWidgetProvider for context management

FLY: Work on integration is still in progress.


src/providers/DepositWidgetProvider.tsx
```diff
@@ -0,0 +1,18 @@
+import type React from "react"
+import type { PropsWithChildren } from "react"
+
+import { ModalContainer } from "../components/Modal/ModalContainer"
+
+import { ModalStoreProvider } from "./ModalStoreProvider"
+import "../styles/main.css"
```
cawabunga-bytes, line 7: Global styles should've been already loaded.

src/features/deposit/components/Form/DepositFormNetwork/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+export const DepositFormNetwork = ({
+  options,
+  onSubmit,
+}: DepositFormNetworkProps) => {
+  const {
+    handleSubmit,
+    register,
+    watch,
+    formState: { errors },
+  } = useForm<DepositFormNetworkValues>({ reValidateMode: "onSubmit" })
```
cawabunga-bytes, line 32: I don't think we need form here, because we don't do anything with form (validate, submit, store data in fields) we can simplify it and just store blockchain in useState.

src/features/deposit/components/DepositFormController.tsx
```diff
@@ -0,0 +1,52 @@
...
+            switch (values.blockchain) {
+              case BlockchainEnum.NEAR:
+                onSelect(DepositFormType.DEPOSIT_NEAR)
+                break
+              case BlockchainEnum.ETHEREUM:
+              case BlockchainEnum.BASE:
+                onSelect(DepositFormType.DEPOSIT_PASSIVE)
+                break
+              default:
+                return null
```
cawabunga-bytes, line 44: Personally I don't like suppressing issues, I would throw here an error, so I would see incorrect data (or incorrect assumption about data) right away.

cawabunga-bytes (approved):
Good work, I left minor comments only

src/features/deposit/components/Form/DepositFormNetwork/index.tsx
jobotics: Well we do submit and store select in a filed, so form my point of view it is still a form, even without validation. The key point is that it uses `reValidateMode` on submit, which ensures that if a user navigates back from a sibling form, we don’t need to worry about resetting the useState constant.

PR #40, opened 2024-10-16
feat(deposit): add passive deposit form

jobotics:
This PR includes the following changes:

- Rename `NetworkTokenSelector` component to `DepositFormRouter`
- Add a custom hook to manage modal payload
- Implement QR code generator for deposit addresses
- Add NEAR whitelist token addresses to fetch balances
- Implement `generateDepositAddress` function in the deposit service
- Update deposit address generation state machine
- Refactor `DepositFormController` select behavior
- Enhance combo input component

Key improvements:
- Improved user experience for deposit address generation
- Better organization of deposit-related components


src/features/deposit/components/DepositWidget.tsx
cawabunga-bytes, line 1: I worry that the purpose of this component is kind of blurry. It basically handles both deposit cases (it implements logic and stores data), so no separation of concerns. Usage of react-hook-form and xstate does make code easier to understand in such case.

src/constants/tokens.ts
```diff
@@ -130,3 +169,133 @@ export const W_BASE_TOKEN_META = {
   decimals: 18,
   routes: [],
 }
+
+export const NEAR_WHITELIST_TOKEN_ADDRESSES = [
```
cawabunga-bytes, line 173: I think we need to move static data to json files.

src/components/Select/Select.tsx
```diff
@@ -47,13 +50,15 @@ export const Select = <T extends string, TFieldValues extends FieldValues>({
         }
         onChange(event)
       }}
+      value={value?.label}
```
cawabunga-bytes, line 53: I don't get this, could you say why `value` is set to `value.label`?

src/components/Form/FieldComboInput.tsx
```diff
@@ -114,6 +116,8 @@ export const FieldComboInput = <T extends FieldValues>({
         !price && !balance && errors && !errors[fieldName] && "pb-5",
         className && className
       )}
+      onClick={handleClick}
+      onKeyUp={handleClick}
```
cawabunga-bytes, line 120: This doesn't make sense for me. Could you explain why we need click handler on non-interactive element?

src/features/deposit/components/DepositWidget.tsx
```diff
@@ -74,12 +96,42 @@ export const DepositWidget = ({
...
+          setFormType(values.formType)
+          setAsset({
+            address: values.address,
+            decimals: values.decimals,
+            icon: values.icon,
+            symbol: values.symbol,
+          })
+          setBlockchain(values.blockchain)
+          if (values.formType === DepositFormType.DEPOSIT_PASSIVE) {
+            generateAddressActor.start()
```
cawabunga-bytes, line 109: Can we automatically start the actor (using hook or special xstate context)?

src/features/deposit/components/DepositWidget.tsx
```diff
@@ -74,12 +96,42 @@ export const DepositWidget = ({
...
+              onBack={() => setFormType(null)}
+            />
+          )}
+        {formType === DepositFormType.DEPOSIT_NEAR && asset && (
           <FormProvider {...depositFormNearMethods}>
-            <DepositFormNear onSubmit={handleSubmitNear} />
+            <DepositFormNear
+              asset={asset}
+              onSubmit={handleSubmitNear}
+              onBack={() => setFormType(null)}
```
cawabunga-bytes, line 133: Don't we need to reset all other data?

src/features/deposit/components/Form/DepositFormRouter/styles.module.css
```diff
@@ -52,3 +57,11 @@
 .orangeButton:hover {
   background-color: darkorange;
 }
+
+.clickableDisabled {
+  cursor: pointer;
```
cawabunga-bytes, line 62: If clicking is disabled then it should be `default` not `pointer`?

src/components/Form/FieldComboInput.tsx
cawabunga-bytes: It does not actually bring trouble. Overall having `onClick` on `div` is incorrect in most cases and biome emphasize it.

src/features/deposit/components/DepositWidget.tsx
jobotics: I don't think that I'm in line with you. Adding a new machine like a UI-machine brings additional complexity and I believe it would be even more complicated.

CC: I suggest we return to this point when we have time for refactoring.

cawabunga-bytes: Ops, I meant

 Usage of react-hook-form and xstate does **not** make code easier to understand in such case.

cawabunga-bytes: When I think about this kind of stuff (like whether component is bloated), then I think how would I unit test this component. And I don't know to how write unit test for this component, so it is a sign for me that something is wrong here.

jobotics: We will have separate tests for both machines, why do we need to test this component as this is just a router between functionality and isn't actual functionality? Even with actors' implementations.

src/components/Select/Select.tsx
jobotics: This is for applying a value indirect, I'll refactor it.

PR #46, opened 2024-10-19
feat: add unified token

cawabunga-bytes:
# Summary

PR introduces `UnifiedTokenInfo` type, which solves natively multichain assets problem. 

```ts
/**
 * A virtual aggregation of the same token across multiple blockchains.
 * This is not an on-chain token but a unified view of network-specific tokens
 * with shared properties.
 *
 * The name avoids "NativeMultichainAsset" to clarify that it doesn't represent
 * an actual multichain token, just a virtual abstraction.
 */
export interface UnifiedTokenInfo {
  unifiedAssetId: string
  symbol: string
  name: string
  decimals: number
  icon: string
  groupedTokens: BaseTokenInfo[]
}
```

<details><summary>Example of definition of USDC</summary>
<p>

```ts
{
    unifiedAssetId: "usdc",
    symbol: "USDC",
    name: "USDC",
    decimals: 6,
    icon: "https://assets.coingecko.com/coins/images/6319/standard/usdc.png",
    groupedTokens: [
      {
        defuseAssetId:
          "near:mainnet:17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1",
        address:
          "17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1",
        decimals: 6,
        icon: "https://assets.coingecko.com/coins/images/6319/standard/usdc.png",
        chainId: "mainnet",
        chainIcon: "/static/icons/network/near.svg",
        chainName: "NEAR",
        routes: [],
        symbol: "USDC",
        name: "USDC",
      },
      // and USDCs on other networks
    ],
  }
```

</p>
</details> 

# Changes
- add `UnifiedTokenInfo` types and type-guards to distinguish `UnifiedTokenInfo` and `BaseTokenInfo`
- update token select and modal to support new type of token

# Motivation

Why `UnifiedTokenInfo`, but not `NativelyMultichainAsset`? Because is just a virtual aggregation, it is not a real token or asset. In the future we may have an actual token which will represent multiple tokens on different chains. But for now it is just FE abstraction.

# Screenshots

<details><summary>Select UI</summary>
<p>

Screenshot 2024-10-19 at 16 48 51

Screenshot 2024-10-19 at 16 48 57

</p>
</details>

### DIFF
--- a/src/components/Modal/ModalSelectAssets.tsx
+++ b/src/components/Modal/ModalSelectAssets.tsx
@@ -2,13 +2,18 @@ import { Text } from "@radix-ui/themes"
 import React, { useState, useDeferredValue, useEffect } from "react"
 import { useModalStore } from "../../providers/ModalStoreProvider"
 import { useTokensStore } from "../../providers/TokensStoreProvider"
-import type { ModalType } from "../../stores/modalStore"
-import type { BaseTokenBalance, BaseTokenInfo } from "../../types/base"
+import { ModalType } from "../../stores/modalStore"
+import type {
+  BaseTokenBalance,
+  BaseTokenInfo,
+  UnifiedTokenInfo,
+} from "../../types/base"
+import { isBaseToken } from "../../utils"
 import { AssetList } from "../Asset/AssetList"
 import { SearchBar } from "../SearchBar"
 import { ModalDialog } from "./ModalDialog"
 
-type Token = BaseTokenInfo
+type Token = BaseTokenInfo | UnifiedTokenInfo
 
 export type ModalSelectAssetsPayload = {
   modalType?: ModalType.MODAL_SELECT_ASSETS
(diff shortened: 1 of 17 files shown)


src/components/Modal/ModalSelectAssets.tsx
```diff
@@ -35,51 +41,83 @@ export const ModalSelectAssets = () => {
...
 
-      if (value?.balance) {
-        getAssetListWithBalances.push({
-          ...value,
-          balance: value?.balance,
-          isNotSelectable,
-        })
+      if (isBaseToken(token)) {
+        if (
+          token.balance != null &&
```
jobotics, line 91: This if condition is incorrect because balance and balanceUsd (calculated using convertedLast, the token rate) represent different features. Even if we successfully fetch the token balance (used for input/output amounts in the swap, deposit, or withdraw forms), the API request to get the market values (for USD conversion) could still fail. Therefore, displaying the token balance should not depend on the market value API call—they should function independently.

cawabunga-bytes: This type confused me:
```ts
type BaseTokenBalance = {
  balance: string
  balanceUsd: string
  convertedLast: {
    [key in BaseTokenConvertEnum]: string
  }
}
```

src/components/Modal/ModalSelectAssets.tsx
```diff
@@ -35,51 +41,83 @@ export const ModalSelectAssets = () => {
...
-  const handleSelectToken = (token: BaseTokenInfo) => {
-    onCloseModal({
-      ...(payload as { fieldName: string }),
-      modalType,
-      token,
-    })
+    const newPayload: ModalSelectAssetsPayload = {
+      ...(payload as ModalSelectAssetsPayload),
+      modalType: ModalType.MODAL_SELECT_ASSETS,
+      token: token.token,
```
jobotics, line 62: The function property `token` seems to have some issues in the token interface. I would suggest renaming it to something like`token: selectedItem.token`.

src/components/Network/AssetList.tsx
```diff
@@ -77,7 +78,7 @@ const AssetList = ({
             type={"button"}
             className={clsx(
               "flex justify-between items-center gap-3 p-2.5 rounded-md hover:bg-gray-950 dark:hover:bg-black-950",
-              isNotSelectable && "opacity-50 pointer-events-none"
```
jobotics, line 80: We should start rewriting all Tailwind styles to CSS Modules to avoid Tailwind collisions between the SDK and the frontend.

src/components/Modal/ModalSelectAssets.tsx
jobotics: You are right, probably we need update it.

jobotics (approved):

PR #51, opened 2024-10-22
feat(deposit): update markup QR code generation for deposit addresses

jobotics:
The PR includes:
- Update DepositFormGenerateAddress component
- Improve UI with instructions and network selection hint
- Update deposit form router
- Add form input component
- Refactor asset combo icon component


src/components/Input/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onChange,
+  className,
+  slotLeft = null,
+  slotRight = null,
+  type = "text",
+}: Props<T, TFieldValues>) => {
+  const registerProps = register ? register(name as Path<TFieldValues>) : {}
+  const { onChange: formOnChange, ...rest } = registerProps as {
+    onChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void
+  }
```
cawabunga-bytes, line 36: We have the same issue here, we should not bind our inputs to `react-hook-form`. I actually want to get rid of react-hook-form from swap widget, it doesn't bring anything to it.

src/components/Input/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <TextField.Root
+      className={clsx(styles.inputTrigger, className, {
+        [styles.inputTriggerFullWidth || ""]: fullWidth,
+        [styles.inputTriggerDisabled || ""]: disabled,
```
cawabunga-bytes, line 41: Oh I know this issue :) Maybe now there is a solution, but long time ago there wasn't so we had to accept no type checker for class names.

cawabunga-bytes (approved):
I would appreciate screenshots next time. Good work! The only important comment is removing unnecessary (`register`) props from input, you're not using after all.

PR #58, opened 2024-10-23
fix: fix token select in swap widget

cawabunga-bytes:


jobotics:
Nice solution with `TokenListUpdater`!

jobotics (approved):

PR #60, opened 2024-10-24
feat: verify public key after singing swap message 

cawabunga-bytes:
# Background
Public key is important part of Defuse protocol. Most blockchain accounts one way or another have public keys. In most blockchains accounts have a single immutable pub key, but on Near accounts can manage their public keys (add or delete them). 

When user signs a message, they use a public key. In order to allow a particular public key to manage user's assets, user needs to tell us about that key by sending a transaction (`add_public_key` to defuse contract). 

# Changes
PR adds an extra step before publishing an intent. On that step it checks presence of the public key (extracted from signature), and if not it urges the user to send a `add_public_key` transaction.

### DIFF
--- a/src/features/machines/publicKeyVerifierMachine.ts
+++ b/src/features/machines/publicKeyVerifierMachine.ts
@@ -1,0 +1,111 @@
+import type { providers } from "near-api-js"
+import type { CodeResult } from "near-api-js/lib/providers/provider"
+import { fromPromise } from "xstate"
+import { settings } from "../../config/settings"
+import type { Transaction } from "../../types/deposit"
+
+export type SendNearTransaction = (
+  tx: Transaction
+) => Promise<{ txHash: string } | null>
+
+export const publicKeyVerifierMachine = fromPromise(
+  async ({
+    input: { nearAccount, nearClient, sendNearTransaction },
+  }: {
+    input: {
+      nearAccount: { accountId: string; publicKey: string } | null
+      nearClient: providers.Provider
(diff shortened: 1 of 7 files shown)


src/features/machines/publicKeyVerifierMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+import type { CodeResult } from "near-api-js/lib/providers/provider"
+import { fromPromise } from "xstate"
+import { settings } from "../../config/settings"
+import type { Transaction } from "../../types/deposit"
+
+export type SendNearTransaction = (
+  tx: Transaction
+) => Promise<{ txHash: string } | null>
+
+export const publicKeyVerifierMachine = fromPromise(
```
jobotics, line 11: This isn't a machine per se, but rather an actor implementation. Let's avoid mixing concerns and create an actors folder within the machines directory, as this is confusing to me and I believe it will confuse others as well.

cawabunga-bytes: It is not mixing concerns. It is co-location, related logic live in one place. They should not be used separately.

jobotics: For now, this isn't a machine because you're only using fromPromise, which is actor logic. It's more like a single piece of logic rather than a complete machine."

cawabunga-bytes: We literally have the same approach in src/features/machines/queryQuoteMachine.ts

PR #61, opened 2024-10-24
refactor: deposit token list

jobotics:
PR includes:
- Removed ModalDepositSelectAssets, as the general modal for token selection is sufficient, avoiding corner cases with differences between swap and deposit asset lists
- Minor props refactoring

Additional changes:
- Extended the settings interface with a blockchains field required for deposits
- Extended the token list interface with a blockchain field, as it's more precise than chainId


src/features/deposit/components/DepositWidget.tsx
```diff
@@ -1,10 +1,11 @@
 import { useState } from "react"
 import { FormProvider, useForm } from "react-hook-form"
+import { assert } from "vitest"
```
cawabunga-bytes, line 3: We should not import vitest in real app.

src/features/deposit/components/DepositWidget.tsx
```diff
@@ -17,23 +18,22 @@ import {
...
   accountId,
-  signAndSendTransactionsNear,
+  sendTransaction,
 }: DepositWidgetProps) => {
   const [formType, setFormType] = useState<DepositFormType | null>(null)
   const [blockchain, setBlockchain] = useState<BlockchainEnum | null>(null)
   const [asset, setAsset] = useState<BaseAssetInfo | null>(null)
   const depositFormNearMethods = useForm<DepositFormNearValues>()
 
+  assert(tokenList.length > 0, "Token list must have at least 1 token")
```
cawabunga-bytes, line 31: What is the purpose of this check?

src/config/settings.ts
```diff
@@ -1,9 +1,24 @@
...
+    },
+    ethereum: {
+      name: BlockchainEnum.ETHEREUM,
+      icon: "/static/icons/network/ethereum.svg",
+    },
+    base: {
+      name: BlockchainEnum.BASE,
+      icon: "/static/icons/network/base.svg",
+    },
+  },
```
cawabunga-bytes, line 21: I don't it is "settings", it is just a list of blockchains.

src/features/deposit/components/DepositWidget.tsx
```diff
@@ -17,23 +18,22 @@ import {
   type DepositFormNearValues,
 } from "./Form/DepositFormNear"
 
-type DepositWidgetProps = {
-  accountId: string
-  signAndSendTransactionsNear: (transactions: Transaction[]) => void
-}
-
```
cawabunga-bytes, line 24: I strong believer that Props should live along with a component itself.

src/features/deposit/components/Form/DepositFormNear/index.tsx
```diff
@@ -24,15 +24,15 @@ export type DepositFormNearValues = {
 
 export interface DepositFormNearProps {
   asset: BaseAssetInfo
-  signAndSendTransactionsNear: (transactions: Transaction[]) => void
+  sendTransaction: (transactions: Transaction[]) => void
```
cawabunga-bytes, line 27: We should be explicit here that we send "Near" transaction. Say we need to implement sending funds directly from the widget, then most likely we will have a bunch of methods like `sendTransactionNear` or `sendTransactionSolana`.

Also it sendTransaction**s** because it accept list of transactions.

And return type might be also incorrect.

src/features/deposit/components/Form/DepositFormNear/index.tsx
```diff
@@ -81,13 +81,14 @@ export const DepositFormNear = ({
 
   useEffect(() => {
     if (asset) {
-      const tokenAdapter: BaseTokenInfo = {
+      const tokenAdapter: SwappableToken = {
```
cawabunga-bytes, line 84: `SwappableToken` belongs to swap widget, I don't recommend using it elsewhere

src/features/deposit/components/Form/DepositFormRouter/index.tsx
```diff
@@ -180,3 +166,40 @@ export const DepositFormRouter = ({ onSubmit }: DepositFormRouterProps) => {
     </Form>
   )
 }
+
+function filterGroupedTokens(
+  token: SwappableToken,
+  blockchain: BlockchainEnum
+) {
+  if (!blockchain) return true
```
cawabunga-bytes, line 174: `blockchain` cannot be falsy

src/features/deposit/components/Form/DepositFormRouter/index.tsx
```diff
@@ -180,3 +166,40 @@ export const DepositFormRouter = ({ onSubmit }: DepositFormRouterProps) => {
...
+  if ("address" in token) {
```
cawabunga-bytes, line 175: We have function `isBaseToken`

src/types/base.ts
```diff
@@ -17,6 +17,7 @@ export interface BaseTokenInfo extends Partial<BaseTokenBalance> {
   name: string
   decimals: number
   icon: string
+  blockchain: string
```
cawabunga-bytes, line 20: We have chainId and chainName. Can we use them instead of adding new property?

src/features/deposit/components/DepositWidget.tsx
jobotics: To deposit we need at least one token in the list, otherwise it won't be able to do anything.

src/config/settings.ts
jobotics: Well, we probably want to set up a deposit network at the SDK level, and maybe not just at the deposit level.

src/features/deposit/components/Form/DepositFormNear/index.tsx
jobotics: Return type isn't finished yet.

src/features/deposit/components/Form/DepositFormNear/index.tsx
jobotics: No, we have a predefined list of tokens, which will be extended later on. From the perspective of the token list, this must remain the same; the only differences could be in the internal identifiers.

src/features/deposit/components/Form/DepositFormRouter/index.tsx
jobotics: Yes, in groupedTokens asset

src/types/base.ts
jobotics: No, because the chain name could be used as a label and in uppercase. I don't think we need it at all.

src/features/deposit/components/DepositWidget.tsx
cawabunga-bytes: Okay, such check assumes you need it to verify assumption about data, but in your case the list will be just empty. And it won't cause any issues.

src/config/settings.ts
cawabunga-bytes: Then it should be a list of strings then.

src/features/deposit/components/Form/DepositFormNear/index.tsx
cawabunga-bytes: Exactly, it was exclusively created for swap widget. If your type has the same value then it is just nice coincedence.

src/features/deposit/components/DepositWidget.tsx
jobotics: No, because this is the root component, which has a type exported from the SDK, and it is the same as for the swap, so then we should take into account it also.

src/features/deposit/components/Form/DepositFormNear/index.tsx
jobotics: This is the DepositFormNear, so we are already in the NEAR context, there’s no need to be more implicit

src/types/base.ts
cawabunga-bytes: I still don't get why we need 3 fields for describing blockchain.

src/features/deposit/components/Form/DepositFormRouter/index.tsx
cawabunga-bytes: `blockchain: BlockchainEnum` means it contains value of BlockchainEnum and BlockchainEnum does not contain any falsy values. I suppose something is wrong here.

src/features/deposit/components/DepositWidget.tsx
cawabunga-bytes: Yes, all components should have props along with them and not separated. If a package wants to export it, then it should re-export it.

src/features/deposit/components/Form/DepositFormNear/index.tsx
cawabunga-bytes: As for me having different names for one thing within a single feature increases congnitive weight.

PR #64, opened 2024-10-25
feat: status of intent

cawabunga-bytes:
# Summary
The main purpose of the PR is to formalize states of the swap flow and statuses of an intent. The main changes are made in `swapIntentMachine`. The swap flow states are grouped by intent creation outcomes:

- "sad" states (we could not create an intent)
  - `ERR_USER_DIDNT_SIGN`
  - `ERR_SIGNED_DIFFERENT_ACCOUNT`
  - `ERR_CANNOT_VERIFY_PUBLIC_KEY`
  - `ERR_CANNOT_ADD_PUBLIC_KEY`
  - `ERR_CANNOT_PUBLISH_INTENT`

- "happy" states (we managed to create an intent)
  - `SETTLED`
  - `NOT_FOUND_OR_NOT_VALID`
  - `ERR_CANNOT_OBTAIN_INTENT_STATUS`

### DIFF
--- a/src/services/intentService.ts
+++ b/src/services/intentService.ts
@@ -1,0 +1,101 @@
+import type { WalletSignatureResult } from "../types"
+import { prepareSwapSignedData } from "../utils/prepareBroadcastRequest"
+import * as solverRelayClient from "./solverRelayHttpClient"
+import type * as types from "./solverRelayHttpClient/types"
+
+export async function submitIntent(
+  signatureData: WalletSignatureResult,
+  quoteHashes: string[]
+) {
+  // todo: retry on network error
+  const result = await solverRelayClient.publishIntent({
+    signed_data: prepareSwapSignedData(signatureData),
+    quote_hashes: quoteHashes,
+  })
+  // todo: check status, it may be "FAILED"
+  return result.intent_hash
+}
+
+export async function waitForIntentSettlement(
+  signal: AbortSignal,
+  intentHash: string
+) {
+  let attempts = 0
+  const MAX_INVALID_ATTEMPTS = 3 // ~600 ms of waiting
+
+  let lastSeenResult: types.GetStatusResponse["result"] | null = null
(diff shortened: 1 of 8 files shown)


jobotics:
Why `NOT_FOUND_OR_NOT_VALID` and `ERR_CANNOT_OBTAIN_INTENT_STATUS` statuses are considered as happy? Either these statuses were unintentionally added to this group, or the names are not consistent here.

src/services/intentService.ts
jobotics, line 1: To maintain a consistent code style, we need to agree on a standardized approach for writing services within the SDK.

Currently, we’ve implemented [depositService.ts](https://github.com/defuse-protocol/defuse-sdk/blob/feat-get-status/src/features/deposit/services/depositService.ts) as a class-based service. Depending on our agreement, this may or may not be rewritten.

So what technic we shoul use servise as a functions or service as a class.

The main question is whether we should implement services as functions or as classes.

### My thoughts:
- If the SDK is performing stateless operations or focuses on simple, independent utility functions, a file with just functions may be the best choice.
- If the SDK requires state management, complex configurations, or needs inheritance for extendibility, then a class-based service would be more suitable.

cawabunga-bytes:
> Why `NOT_FOUND_OR_NOT_VALID` and `ERR_CANNOT_OBTAIN_INTENT_STATUS` statuses are considered as happy? Either these statuses were unintentionally added to this group, or the names are not consistent here.

You can call it either name you want, "happy" means we managed to created a intent.

src/services/intentService.ts
cawabunga-bytes: Yeah, it makes sense. Just functions in frontend world are better when it comes to tree-shaking. Also I'm not sure if we will have any classes (methods + state), because we store either in react hooks or in xstate machines.

jobotics (approved):

PR #65, opened 2024-10-25
refactor: separate intent

cawabunga-bytes:
# Summary

This PR is follow-up of https://github.com/defuse-protocol/defuse-sdk/pull/64 
The main goal of this PR is to separate intent statuses from `swapIntentMachine`.

`swapIntentMachine` will just create an intent and return its hash. `swapUIMachine` will grab the result and create another actor which will be responsible for storing the status of the intent and poll API for a settlement status.

Such separation enables granular control over the intents, so we could render them easily concurrently.

<details><summary>Screenshot: How we can render multiple concurrent intents</summary>
<p>

Screenshot 2024-10-25 at 20 21 43

</p>
</details>

### DIFF
--- a/src/features/swap/components/SwapForm.tsx
+++ b/src/features/swap/components/SwapForm.tsx
@@ -1,6 +1,7 @@
 import { useSelector } from "@xstate/react"
-import { useContext, useEffect } from "react"
+import { Fragment, useContext, useEffect } from "react"
 import { useFormContext } from "react-hook-form"
+import { formatUnits } from "viem"
 import type { ActorRefFrom } from "xstate"
 import { ButtonCustom, ButtonSwitch } from "../../../components/Button"
 import { Form } from "../../../components/Form"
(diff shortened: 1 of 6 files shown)


src/features/swap/components/SwapForm.tsx
```diff
@@ -119,46 +123,119 @@ export const SwapForm = () => {
...
+          {swapInfo} 💤
+          <br />
+          Checking intent status... intentHash: {snapshot.context.intentHash}{" "}
+          tx: {snapshot.context.txHash}
         </div>
       )
+    case "success":
+      return (
+        <div>
+          {swapInfo} ✅
```
jobotics, line 187: Why are we using emoji icons? If we’re missing any icons, we should request them from the designer.

cawabunga-bytes: This is temporary, because we don't have mockups yet, because design was requested this Friday, and you were on the meeting as well :)

src/features/swap/components/SwapForm.tsx
```diff
@@ -119,46 +123,119 @@ export const SwapForm = () => {
...
+      <Intents intentRefs={snapshot.context.intentRefs} />
     </div>
   )
 }
 
-function renderIntentOutcome(intentOutcome: Context["outcome"]) {
-  if (!intentOutcome) {
-    return null
-  }
+function Intents({
```
jobotics, line 132: `Intents` should be a separate component to avoid overloading the SwapForm itself.

cawabunga: Yeah, it will most likely be in own file. I left it here because it's still in progress and just reflects a possibility to shows statuses of intents.

cawabunga-bytes: Yeah, it will be most likely migrated to own file, as soon as design will be ready. Now it is just an example of what we can do with intents.

PR #66, opened 2024-10-26
feat: re-quote during swap

cawabunga-bytes:
# Summary

PR extracts re-querying quotes from `swapUIMachine`. Now quotes are re-queried constantly in the background and then the new quote is propagated to states. 

If a quote is updated
1. During editing phase, it is just updates the amount out field.
2. During swapping phase, it compares the active quote and replace it, if the new quote is better.


src/features/machines/backgroundQuoterMachine.ts
jobotics, line 1: Once again, I recommend rethinking the current pattern for state machines and actors. This is not a fully-fledged state machine but rather a hybrid actor.

## Why this is problematic
This approach results in a bloated `machines` folder, filled with files that are not true state machines. We could end up with a structure containing 5 actual state machines and 95 files with actors (functions that these machines rely on). 

## Solution
I suggest creating a dedicated `actors` folder inside  `machines` folder for all actor files, and if an actor evolves into a true state machine with setup and createMachine, then it can be moved to the machines folder.

cawabunga-bytes: Alright, I still don't get what a true state machine is, but if you think it will be better for navigation then we can move files of course.

jobotics (approved):

PR #67, opened 2024-10-26
feat: defuse tokens balances

cawabunga-bytes:
# Summary
PR adds support for balances of the user within Defuse Protocol. As an example, I added printing balances under inputs.

# Technical Details
PR introduces new depositedBalance machine, which responsible for querying and storing balance. All communication between sub-systems happen through root `swapUIWidget` machine, it just relays events to proper actor. For example, a scenario:
1. User connects a wallet
2. useEffect tracks `userAddress` and sends `LOGIN` event to `swapUIMachine`
3. `swapUIWidget` relays `LOGIN` event to `depositedBalanceMachine`
4. `depositedBalanceMachine` fetches tokens from defuse smart-contract
5. `depositedBalanceMachine` sends `BALANCE_CHANGED` to `swapUIMachine`
6. `swapUIMachine` relays `BALANCE_CHANGED` to `backgroundQuoter`
7. `backgroundQuoter` re-queries the quote if input is valid

# Follow-up

- Need to provide a token list to the machine, as for now it just fetches balances of USDT and wNEAR. https://github.com/defuse-protocol/defuse-sdk/pull/68
- Render balances in token select dialog https://github.com/defuse-protocol/defuse-sdk/pull/69


jobotics:
How can the `depositedBalance` machine be reused on the portfolio page? We should prepare it for such scenarios.

cawabunga-bytes:
> How can the `depositedBalance` machine be reused on the portfolio page? We should prepare it for such scenarios.

It could be used directly without issues, I wrote it bearing in mind such case. But of course I couldn't tell you for sure, because I don't know all use-cases and access patterns in portfolio.

jobotics (approved):

PR #62, opened 2024-10-25
feat(deposit): implement transaction batching for NEAR deposits

jobotics:
## Changes NEAR Deposit Enhancements
- Added support for batching transactions when depositing native NEAR as collateral for wNEAR
- Updated depositNearMachine to handle multi-transaction scenarios
- Implemented createNativeDepositNearTransaction for native NEAR deposits
- Enhanced DepositFormNear component to utilize new batching functionality

## Transaction Validation and Status
- Added checkNearTransactionValidity function for transaction verification
- Updated depositNearMachine with transaction validation logic
- Implemented getTxStatus for transaction status retrieval

## Blockchain Interactions and State Management
- Added nearHttpClient for NEAR blockchain interactions
- Introduced RoopProvider for managing application state
- Created getBalanceMachine for balance fetching logic
- Implemented RTK query for optimized API requests

## UI and User Management
- Updated AssetList component to display AssetBalance
- Added UserProvider for user-related context management

These changes significantly improve the NEAR deposit process, enhance transaction validation, and optimize state management and API interactions throughout the application.

CC: Sorry for the tons of updates.

PR #71, opened 2024-10-28
fix: improve deposit flow

jobotics:
- Fix batch transaction processing to ensure atomicity and consistency.
- Implement semaphore to mitigate rate limit requests, enhancing reliability.
- Refactor asset list to remove redundant structures.
- Fix get balance machine to accurately reflect current balances and handle edge cases.

CC: Included correction for #62


src/components/Asset/AssetList.tsx
```diff
@@ -51,11 +54,24 @@ export const AssetList = <T extends Token>({
...
+  const { data: balances, mutate: fetchBalances } = useGetNearBalances()
+
+  useEffect(() => {
+    if (accountId) {
+      fetchBalances({
+        tokenList: assets.map((item) => item.token),
+        userAddress: accountId,
+      })
+    }
+  }, [accountId, fetchBalances, assets])
```
cawabunga-bytes, line 73: Use can granularly control how useQuery works, for example, you can use `enabled` to disable running the request:
```js
useQuery({..., enabled: accountId != null })
```

src/components/Asset/AssetList.tsx
```diff
@@ -110,3 +132,37 @@ export const AssetList = <T extends Token>({
...
+    if (isGroupedToken) {
+      return tokenIds.reduce((acc, assetId) => {
+        const balance = formatUnits(balances[assetId] ?? BigInt(0), decimals)
+        return (Number(acc) + Number(balance)).toString()
+      }, "0")
+    }
+    if (defuseAssetId === "nep141:wrap.near") {
+      return formatUnits(
+        (balances[defuseAssetId] ?? BigInt(0)) +
+          (balances["near:native"] ?? BigInt(0)),
```
cawabunga-bytes, line 160: Instead of storing not real defuse token (`near:native`) in balances, might be better to store a balance in separate variable. In that way we will be 100% sure that `defuseAssetId` is always token id within defuse protocol.

src/components/Modal/ModalSelectAssets.tsx
```diff
@@ -80,43 +82,30 @@ export const ModalSelectAssets = () => {
...
-              balanceUsd: token.balanceUsd,
-              convertedLast: token.convertedLast,
-            },
-          })
-        }
-      } else {
-        const hasBalance = token.groupedTokens.some(
-          (innerToken) => innerToken.balance != null
+      if (isUnifiedToken(token)) {
+        const defuseAssetId = token.groupedTokens.reduce(
```
cawabunga-bytes, line 89: I didn't get why need this, even after reading the comment :) If it is needed for sure then you need to consider a better for it, because it is not `defuseAssetId`.

cawabunga-bytes (approved):
Good work! I left some comments. Feel free to merge as soon as you think it's ready.

src/features/machines/getBalanceMachine.ts
```diff
@@ -51,3 +55,65 @@ export const getNearNep141Balance = async ({
...
+          }
+        } finally {
+          semaphore.release()
+        }
+      }),
+      (async () => {
+        await semaphore.acquire()
+        try {
+          return {
+            "near:native": await getNearNativeBalance({ userAddress }),
```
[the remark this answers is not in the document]
cawabunga-bytes: This is not relevant, because it will be a separate variable.

src/utils/semaphore.ts
```diff
@@ -0,0 +1,33 @@
...
+  release(): void {
+    this.currentRequests--
+
+    if (this.queue.length > 0) {
+      const next = this.queue.shift()
+      setTimeout(() => {
+        this.currentRequests++
+        // @ts-ignore: next is guaranteed to be defined here
+        next()
+      }, this.delayMs)
```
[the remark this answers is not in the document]
jobotics: A few reasons:
1. Rate Limiting: By introducing a delay, you can control the rate at which requests are processed. This can be useful if you're interacting with an external service that has rate limits.
2. Preventing Starvation: If the semaphore is used in a high-concurrency environment, the delay can help prevent starvation of queued tasks by ensuring that they are given a chance to execute in a controlled manner.
3. Asynchronous Behavior: The setTimeout function ensures that the queued function is executed asynchronously, allowing the event loop to process other tasks in the meantime. This can be beneficial in a non-blocking environment like Node.js, where you want to keep the event loop free to handle other I/O operations.

src/components/Modal/ModalSelectAssets.tsx
jobotics: For example, we need this on the deposit page, where we can't use the mt_balance call to retrieve a multi-asset grouped balance. Instead, we need to fetch balances by specific `defuseAssetIds`, such as USDC and ETH, rather than by `unifiedAssetId`.

cawabunga-bytes: Won't be easier to pass an array instead of serializing to string and then deserializing?

src/utils/semaphore.ts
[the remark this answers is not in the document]
cawabunga-bytes: Rate limiting should not be part of "semaphore", it is just a simple counter. If you want to have it asynchronous then you can use `queueMicrotask`.

src/components/Modal/ModalSelectAssets.tsx
jobotics: None of these options feel ideal to me. I’m inclined to leave it for now and revisit it when we have more time.

PR #74, opened 2024-10-28
fix: query quote each 500ms

cawabunga-bytes:
# Summary

PR make quotes requests send each 500 ms regardless of anything. But only the latest seen quote is taken, all other quotes will be discarded.

### DIFF
--- a/src/features/machines/backgroundQuoterMachine.ts
+++ b/src/features/machines/backgroundQuoterMachine.ts
@@ -1,4 +1,5 @@
 import { type ActorRef, type Snapshot, fromCallback } from "xstate"
+import { settings } from "../../config/settings"
 import type { BaseTokenInfo, UnifiedTokenInfo } from "../../types/base"
 import { isBaseToken } from "../../utils"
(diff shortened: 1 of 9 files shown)


src/features/machines/backgroundQuoterMachine.ts
```diff
@@ -85,27 +92,64 @@ async function pollQuoteLoop(
...
-      return null
-    })
+    const requestedAt = Date.now()
 
-    if (quote == null || signal.aborted) {
-      return
-    }
+    queryQuote(
+      {
+        tokensIn: getUnderlyingDefuseAssetIds(quoteInput.tokenIn),
```
jobotics, line 102: Looks like it’s more accurate to say this property is `tokenInIds` rather than a `tokensIn` object.

cawabunga-bytes: Yeah, I also thought about it, but I didn't like separation of words "token" and "id", so my second thought was to have `tokenIdsIn` but it looked even worse, so I decided to go with just `tokensIn`

PR #75, opened 2024-10-28
fix: update style of swap intent

cawabunga-bytes:
It is not final design, but at least it looks prettier and doesn't look broken.

<p>

Screenshot 2024-10-28 at 22 06 52

</p>
</details>


jobotics:
Are we planning to leave it as-is, or is this a draft design? I’m also curious about how we’ll display error events.

jobotics (approved):

cawabunga-bytes:
> Are we planning to leave it as-is, or is this a draft design? I’m also curious about how we’ll display error events.

I don't think this is final design. Errors are expected to look like this:

Screenshot 2024-10-29 at 13 22 24

PR #84, opened 2024-10-30
chore(deposit): disable network for non muli-asset

jobotics:
### Prerequisites:
When the user is handling NEAR tokens, the network is already known, so there’s no need to prompt the user to select it manually; instead, selection is disabled and handled automatically.

Changes:
- Adjust the selection behavior, as the form may be modified not only by the user

### DIFF
--- a/src/features/deposit/components/DepositForm/index.tsx
+++ b/src/features/deposit/components/DepositForm/index.tsx
@@ -21,7 +21,7 @@ import { useModalStore } from "../../../../providers/ModalStoreProvider"
 import { ModalType } from "../../../../stores/modalStore"
 import { DepositBlockchainEnum, type SwappableToken } from "../../../../types"
 import { formatTokenValue } from "../../../../utils/format"
-import { isBaseToken } from "../../../../utils/token"
+import { isBaseToken, isUnifiedToken } from "../../../../utils/token"
 import { DepositUIMachineContext } from "../DepositUIMachineProvider"
 import styles from "./styles.module.css"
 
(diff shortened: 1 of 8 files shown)


src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -323,3 +334,11 @@ function getBlockchainsOptions(): Record<
 function shortenTxHash(txHash: string) {
   return `${txHash.slice(0, 5)}...${txHash.slice(-5)}`
 }
+
+function getDefaultBlockchainOption(token: SwappableToken) {
+  if (token && !isUnifiedToken(token)) {
+    const key = `${token.chainName}:${token.chainId}`
```
cawabunga-bytes, line 340: This is important format, so I suggest leaving a comment about wtf this format means. Also suggest not using `chainId` at all.

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -323,3 +334,11 @@ function getBlockchainsOptions(): Record<
 function shortenTxHash(txHash: string) {
   return `${txHash.slice(0, 5)}...${txHash.slice(-5)}`
 }
+
+function getDefaultBlockchainOption(token: SwappableToken) {
+  if (token && !isUnifiedToken(token)) {
```
cawabunga-bytes, line 339: `token` is always defined.

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -323,3 +334,11 @@ function getBlockchainsOptions(): Record<
...
+    const key = `${token.chainName}:${token.chainId}`
+    return getBlockchainsOptions()[key]
+  }
+  return undefined
```
cawabunga-bytes, line 343: If it is `BaseToken` then the value should be `{token.chainName}:${token.chainId}`, because BaseToken always has blockchain info.

PR #88, opened 2024-11-03
chore(deposit): enhance balance block and input focus

jobotics:
Refactor:
- Provide a ref to the input to allow auto-focus on the deposit amount input
- Enhance styles in `BlockMultiBalances` to allow customization

Changes:
- Add TooltipInfo component
- Fix `isInsufficientBalance` function


src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -131,13 +126,21 @@ export const DepositForm = () => {
...
     network
   )
 
+  const amountInputRef = useRef<HTMLInputElement | null>(null)
+
+  useEffect(() => {
+    if (network === BlockchainEnum.NEAR && amountInputRef.current) {
+      setTimeout(() => amountInputRef.current?.focus(), 0)
+    }
+  }, [network])
```
cawabunga-bytes, line 142: Hard way to auto focus. Here are other options:
1. use `autoFocus` attribute
2. use function ref: `<input ref={(ref) => ref?.focus()} />`

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -427,12 +448,15 @@ function isInsufficientBalance(
...
     if (isBaseToken(token) && token.address === "wrap.near") {
-      return formAmount >= balanceToFormat + nativeBalanceFormatted
+      const balanceToFormat = formatTokenValue(
+        balance + nativeBalance,
+        token.decimals
+      )
+      return formAmount > balanceToFormat
     }
-    return formAmount >= balanceToFormat
+    return formAmount > balanceToFormat
```
cawabunga-bytes, line 459: DIdn't pay to attention to it previously, but you should not compare strings, as you can get unexpected result. For example, `"10" > "9"`will give false.

cawabunga-bytes (approved):

src/features/deposit/components/DepositForm/index.tsx
jobotics: Good catch, thanks!

PR #89, opened 2024-11-03
chore: fetch minimum storage deposit for near nep-141 tokens

jobotics:
## Intentions
Ensure accurate retrieval of minimum storage deposit for each NEAR NEP-141 token to prevent transaction failures for new users.

### DIFF
--- a/src/features/deposit/components/DepositUIMachineProvider.tsx
+++ b/src/features/deposit/components/DepositUIMachineProvider.tsx
@@ -85,12 +85,17 @@ export function DepositUIMachineProvider({
                 if (tokenAddress === "wrap.near") {
                   const isWrapNearRequired =
                     Number(amount) > Number(balance || 0n)
+                  const minStorageBalance =
+                    await depositNearService.getMinimumStorageBalance(
+                      tokenAddress
+                    )
                   transactions =
                     depositNearService.createBatchDepositNearNativeTransaction(
                       tokenAddress,
                       amount,
                       amount - balance,
-                      isWrapNearRequired
+                      isWrapNearRequired,
+                      minStorageBalance
                     )
                 } else {
                   const isStorageDepositRequired =
@@ -98,11 +103,16 @@ export function DepositUIMachineProvider({
                       tokenAddress,
                       settings.defuseContractId
                     )
+                  const minStorageBalance =
+                    await depositNearService.getMinimumStorageBalance(
+                      tokenAddress
+                    )
                   transactions =
                     depositNearService.createBatchDepositNearNep141Transaction(
                       tokenAddress,
                       amount,
-                      isStorageDepositRequired
+                      isStorageDepositRequired,
+                      minStorageBalance
                     )
                 }
 
--- a/src/features/deposit/components/DepositUIMachineProvider.tsx
+++ b/src/features/deposit/components/DepositUIMachineProvider.tsx
@@ -103,10 +103,14 @@ export function DepositUIMachineProvider({
                       tokenAddress,
                       settings.defuseContractId
                     )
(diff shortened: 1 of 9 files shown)


src/features/deposit/services/depositService.ts
```diff
@@ -181,6 +183,12 @@ export class DepositService {
       contractId,
       accountId,
     })
-    return storageBalance < BigInt(FT_MINIMUM_STORAGE_BALANCE)
+    return storageBalance < (await this.getMinimumStorageBalance(contractId))
+  }
+
+  async getMinimumStorageBalance(contractId: string): Promise<bigint> {
+    return await getNearNep141MinStorageBalance({
```
cawabunga-bytes, line 190: nit: `await` does nothing here

src/features/deposit/services/depositService.ts
```diff
@@ -181,6 +183,12 @@ export class DepositService {
       contractId,
       accountId,
     })
-    return storageBalance < BigInt(FT_MINIMUM_STORAGE_BALANCE)
+    return storageBalance < (await this.getMinimumStorageBalance(contractId))
```
cawabunga-bytes, line 186: So you will do 2 requests for the same data, right?

src/features/machines/getBalanceMachine.ts
```diff
@@ -133,3 +134,19 @@ export const getNearNep141StorageBalance = async ({
...
+export const getNearNep141MinStorageBalance = async ({
+  contractId,
+}: {
+  contractId: string
+}): Promise<bigint> => {
+  const response = await getNearNep141StorageBalanceBounds({
+    request_type: "call_function",
+    method_name: "storage_balance_bounds",
+    account_id: contractId,
+    args_base64: "e30=",
```
cawabunga-bytes, line 147: What is this magic? :) Let's write it explicity `base64.encode(JSON.stringify({}))`

src/features/machines/getBalanceMachine.ts
```diff
@@ -133,3 +134,19 @@ export const getNearNep141StorageBalance = async ({
...
+    finality: "optimistic",
+  })
+  const balance = JSON.parse(Buffer.from(response.result).toString())
```
cawabunga-bytes, line 150: I would avoid using Buffer, because browsers don't have it. We can write it like this:
```js
  const uint8Array = new Uint8Array(response.result)
  const decoder = new TextDecoder()
  const parsed = JSON.parse(decoder.decode(uint8Array))
  ```

File: src/features/deposit/components/DepositUIMachineProvider.tsx [hunk 5]
```diff
@@ -85,24 +85,34 @@ export function DepositUIMachineProvider({
...
                 } else {
                   const isStorageDepositRequired =
                     await depositNearService.isStorageDepositRequired(
                       tokenAddress,
                       settings.defuseContractId
                     )
+                  const minStorageBalance =
+                    await depositNearService.getMinimumStorageBalance(
+                      tokenAddress
+                    )
```
cawabunga-bytes, line 109: You can avoid doing this request if `isStorageDepositRequired` is false

cawabunga-bytes (approved):

src/features/deposit/services/depositService.ts
jobotics: I see the confusion. No, we actually make two different requests. The first is `getNearNep141StorageBalance`, which uses the method `storage_balance_of` to tell us the user's deposit. The second is `getNearNep141MinStorageBalance`, which uses the method `storage_balance_bounds` to tell us the minimum required deposit value.

PR #99, opened 2024-11-06
chore: use protocol convention for user address

jobotics:


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -1,5 +1,6 @@
 import { createActorContext } from "@xstate/react"
 import type { PropsWithChildren, ReactElement, ReactNode } from "react"
+import { userAddressToDefuseUserId } from "src/utils/defuse"
```
cawabunga-bytes, line 3: Please replace with relative import.

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -152,7 +153,7 @@ export function DepositUIMachineProvider({
                 assert(accountId != null, "Account ID is not defined")
                 assert(chain != null, "Chain is not defined")
                 const address = await depositNearService.generateDepositAddress(
-                  accountId,
+                  userAddressToDefuseUserId(accountId),
```
cawabunga-bytes, line 156: `depositNearService.generateDepositAddress` should have `DefuseUserId` requirement, so no one could accidentally pass incorrect string.

src/features/deposit/services/depositService.ts
```diff
@@ -126,12 +127,12 @@ export class DepositService {
   /**
    * Generate a deposit address for the specified blockchain and asset through the POA bridge API call.
    *
-   * @param blockchain - The blockchain for which to generate the address
-   * @param accountId - The address of the asset being deposited
+   * @param userAddress - The user address from the wallet
```
cawabunga-bytes, line 130: nit: it's not address from wallet, it's an account within defuse protocol.

cawabunga-bytes (approved):

PR #134, opened 2024-11-12
refactor: rewrite deposit class service to functions

jobotics:
Key changes:
- Move deposit service to a new location and refactor
- Add `getAvailableDepositRoutes` for getting available deposit routes for a given wallet connection and selected network


src/services/depositService.ts
```diff
@@ -0,0 +1,245 @@
...
+        case BlockchainEnum.ETHEREUM:
+        case BlockchainEnum.BASE:
+        case BlockchainEnum.ARBITRUM:
+        case BlockchainEnum.BITCOIN:
+          return {
+            activeDeposit: false,
+            passiveDeposit: true,
+          }
+        default:
+          return null
```
cawabunga-bytes, line 223: Suggest adding exhaustive check, so if you add new blockchain then it would yell here.

src/services/depositService.ts
```diff
@@ -0,0 +1,245 @@
...
+        case BlockchainEnum.NEAR:
+          return {
+            activeDeposit: false, // TODO: Implement
+            passiveDeposit: false,
+          }
+        case BlockchainEnum.ETHEREUM:
+        case BlockchainEnum.BASE:
+        case BlockchainEnum.ARBITRUM:
+          return {
+            activeDeposit: true,
```
cawabunga-bytes, line 236: It's not implemented yet.

jobotics: Correct, on #137. Make it false for now

cawabunga-bytes (approved):

PR #140, opened 2024-11-15
feat: add storage deposit

cawabunga-bytes:


src/constants/tokens.ts
```diff
@@ -155,3 +157,16 @@ export const NEAR_WHITELIST_TOKEN_ADDRESSES = [
   "v1.guild-covenant.near",
   "chill-129.meme-cooking.near",
 ]
+
+export const NEP141_STORAGE_TOKEN: BaseTokenInfo = {
```
jobotics, line 161: With this, I think we no longer need the W_NEAR_TOKEN_META constant above. Let's remove it."

src/features/machines/swapIntentMachine.ts
```diff
@@ -588,3 +605,37 @@ async function verifyWalletSignature(
...
+  nep141Storage: NEP141StorageRequirement | null,
+  directWithdrawalAmount: bigint
+): bigint {
+  const gotFromSwap = swapQuote?.totalAmountOut ?? 0n
+
+  let spentOnStorage = 0n
+  if (nep141Storage != null) {
+    if (nep141Storage.type === "no_swap_needed") {
+      // Assume that token out is NEAR/wNEAR, so we can just use the required storage
+      spentOnStorage = nep141Storage.requiredStorageNEAR
```
jobotics, line 634: I didn't get how you do the calculation of requiredStorageNEAR. Are you getting this from the solver?

src/constants/tokens.ts
cawabunga-bytes: Yeah, probably. We have a lot of code to remove. We'll need to remove it at some point.

src/features/machines/swapIntentMachine.ts
cawabunga-bytes: I get it from chain in `getNEP141StorageRequired`. Then it is used in `determineNEP141StorageRequirement` to know if swap needed or not.

src/features/withdraw/components/WithdrawForm/selectors.ts
jobotics, line 1: I believe selectors.ts is not part of the component itself. Can we create a dedicated folder inside the machine for it? This could provide a clearer architecture for future implementations of such selectors.

PR #142, opened 2024-11-18
refactor: deposit flow enhancement

jobotics:
# Key changes

Use Default RPCs Map: 
- Implemented a default mapping for RPCs to streamline interactions

Remove RPC from External Interface: 
- Refactored the external interface to eliminate direct RPC dependencies, enhancing encapsulation

Refactor Deposit Machine: 
- Updated the deposit machine to rely on the default RPCs, improving maintainability

Clean Up Deposit Form:
- Simplified the responsibilities of the deposit form

[IMPORTANT] Remove `ethers`: 
- Replaced the `ethers` library with `Viem` for better performance and compatibility

Add Estimation Services: 
- Introduced estimation services to provide information on transaction costs and limits


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -56,11 +56,8 @@ export const DepositUIMachineContext: DepositUIMachineContextInterface =
 
 interface DepositUIMachineProviderProps extends PropsWithChildren {
   tokenList: SwappableToken[]
-  sendTransactionNear: (transactions: Transaction[]) => Promise<string | null>
-  sendTransactionEVM: (calldata: {
-    to?: string
-    data?: string
-  }) => Promise<string | null>
+  sendTransactionNear: (tx: Transaction["NEAR"][]) => Promise<Hash | null>
```
cawabunga-bytes, line 59: Near transaction won't return `Hash`. `Hash` is `0x${string}`, near transaction hash has different structure.

src/config/settings.ts
```diff
@@ -22,6 +23,18 @@ export let settings: Settings = {
...
+   * RPC URLs for different blockchains.
+   * Ensure these URLs are valid and accessible.
+   */
+  rpcUrls: {
+    near: "https://nearrpc.aurora.dev" as RpcUrl,
+    ethereum: "https://cloudflare-eth.com" as RpcUrl,
+    base: "https://mainnet.base.org" as RpcUrl,
+    arbitrum: "https://arb1.arbitrum.io/rpc" as RpcUrl,
+    bitcoin: "https://mainnet.bitcoin.org" as RpcUrl,
+    solana: "https://api.mainnet-beta.solana.com" as RpcUrl,
```
cawabunga-bytes, line 36: Let's reuse chain names from `src/types/base.ts:14` and not make up new names for chains, it's becoming overwhelming to memorize what's what.

src/utils/adapters.ts
```diff
@@ -7,3 +7,11 @@ export const assetNetworkAdapter: Record<string, BlockchainEnum> = {
   arbitrum: BlockchainEnum.ARBITRUM,
   bitcoin: BlockchainEnum.BITCOIN,
 }
+
+export const reverseAssetNetworkAdapter: Record<BlockchainEnum, string> = {
```
cawabunga-bytes, line 11: it's not just string, let's use `SupportedChainName`  from `src/types/base.ts:14`

src/utils/defuse.ts
```diff
@@ -12,3 +12,8 @@ export type DefuseUserId = string & { __brand: "DefuseAccountId" }
 export function userAddressToDefuseUserId(userAddress: string): DefuseUserId {
   return userAddress.toLowerCase() as DefuseUserId
 }
+
+// Branding a string type with a unique symbol to prevent accidental misuse.
+// Used in settings.ts to ensure RPC URLs are properly typed and validated.
+// See settings.ts for example usage with different blockchain networks.
+export type RpcUrl = string & { __brand: "RpcUrl" }
```
cawabunga-bytes, line 19: I don't think we need branded string for URL. Branded types should be used when it's critical to use nominal types. But in our case it creates more friction than needed, say we allow users to override RPC URLs, then how would users get this type? They will just typecast (`as RpcUrl`, btw you do the same), which does not help with type safety. I suggest getting rid of it and use just `string`.

Also it's not a `defuse` specific thing, so if you insist that we need this branded string, then it should be moved somewhere else.

src/types/quoting.ts
```diff
@@ -31,4 +33,12 @@ export interface Settings {
   queries: {
     staleTime: number
   }
+  rpcUrls: {
```
cawabunga-bytes, line 36: `Record<SupportedChainName, string>` will be better, so if new chain is added, it will yell at you saying you missed RPC URL here. The same pattern should be used in all other places where chain name is involved.

src/services/depositService.ts
```diff
@@ -277,3 +279,21 @@ export function getAvailableDepositRoutes(
...
+    case BlockchainEnum.ETHEREUM:
+      return settings.rpcUrls.ethereum
+    case BlockchainEnum.BASE:
+      return settings.rpcUrls.base
+    case BlockchainEnum.ARBITRUM:
+      return settings.rpcUrls.arbitrum
+    case BlockchainEnum.BITCOIN:
+      return settings.rpcUrls.bitcoin
+    default:
+      throw new Error("exhaustive check failed")
```
cawabunga-bytes, line 297: You can easily miss Solana here, when it's added.

src/services/estimateService.ts
```diff
@@ -0,0 +1,54 @@
...
+    value,
+  })
+  const gasPrice = await client.getGasPrice()
+  // Add 15% buffer to gas cost estimation to account for potential price fluctuations
+  const costInWei = (gasPrice * gasLimit * 115n) / 100n
+  return costInWei
+}
+
+// Function returns the gas cost in Wei for a transfer
+export async function estimateEVMErc20TransferCost({
```
cawabunga-bytes, line 31: Are these two functions do the same?

src/features/deposit/components/DepositUIMachineProvider.tsx
jobotics: Oh, you are right

src/utils/defuse.ts
jobotics: Agree, cool points!

src/services/estimateService.ts
jobotics: Absolutely, I'll remove it

PR #144, opened 2024-11-18
fix: fix support of solana chain

cawabunga-bytes:


src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
```diff
@@ -38,17 +40,17 @@ export function DepositUIMachineFormSyncProvider({
...
-    if (!userAddress) {
+    if (!userAddress || userChainType == null) {
       actorRef.send({
         type: "LOGOUT",
       })
     } else {
       actorRef.send({
         type: "LOGIN",
-        params: { userAddress, rpcUrl },
+        params: { userAddress, userChainType, rpcUrl },
```
jobotics, line 50: Just note `rpcUrl` no longer exists in the main branch. I think it's worth updating the branch to ensure that the latest PR to the main is smoothly merged.

jobotics: Be careful when pulling from the `main` branch at [src/features/machines/depositUIMachine.ts](https://github.com/defuse-protocol/defuse-sdk/pull/144/files#diff-c08af5996b118d9a146a286a055fce826f7525c4dcdf4af447e8dff5851b31b8). There might be a small conflict.

src/features/machines/depositGenerateAddressMachine.ts
```diff
@@ -7,7 +7,8 @@ export type DepositGeneratedDescription = {
 }
 
 type Context = {
-  accountId: string | null
+  userAddress: string
+  userChainType: ChainType
```
jobotics, line 11: **JMT:** Prefix `user` makes `ChainType` more specific but still I feel some misleading.

jobotics: But when it comes together with `userAddress` it gives me more sense, so all is good.

src/features/machines/depositedBalanceMachine.ts
```diff
@@ -38,14 +42,14 @@ type ThisActor = ActorRef<Snapshot<unknown>, SharedEvents>
...
-  | { type: "LOGIN"; params: { accountId: string } }
+  | { type: "LOGIN"; params: { userAddress: string; userChainType: ChainType } }
 
 export const depositedBalanceMachine = setup({
   types: {
     context: {} as {
       parentRef: ParentActor
       defuseTokenIds: string[]
-      userAccountId: string | null
+      userAccountId: DefuseUserId | null
```
jobotics, line 52: Probably we should rename `userAccountId` to `useAddress`. The same as we do for just `accountId`.

src/features/machines/depositedBalanceMachine.ts
```diff
@@ -83,7 +87,7 @@ export const depositedBalanceMachine = setup({
   },
   actions: {
     setUserAccountId: assign({
```
jobotics, line 89: Same here for suffix `AccountId`, is could be rename to `setUserAddress`.

jobotics (approved):

src/features/machines/depositedBalanceMachine.ts
cawabunga-bytes: Yes, I deliberately skipped all the refactoring part, because it would be hard to review all changes.

PR #146, opened 2024-11-20
feat: add publish intent check

depressedPlumber502:


src/features/machines/swapIntentMachine.ts
```diff
@@ -501,14 +502,37 @@ export const swapIntentMachine = setup({
...
+        onDone: [
+          {
+            target: "Completed",
+            guard: {
+              type: "isOk",
+              params: ({ event }) => event.output,
+            },
+            actions: {
+              type: "setIntentHash",
+              params: ({ event }) => event.output.value as string,
```
cawabunga-bytes, line 514: It's better to assert value instead of typecasting, because guard could be removed for some reason:
```
assert(event.output.tag === 'ok')
return event.output.value
```

src/features/machines/swapIntentMachine.ts
```diff
@@ -501,14 +502,37 @@ export const swapIntentMachine = setup({
...
+            },
           },
-        },
+          {
+            target: "Generic Error",
+            actions: [
+              {
+                type: "logError",
```
cawabunga-bytes, line 521: I doubt we need to log this, as it's not exception, it's something that could happen and we expect it.

src/features/machines/swapIntentMachine.ts
```diff
@@ -501,14 +502,37 @@ export const swapIntentMachine = setup({
...
+                params: ({ event }) => ({
+                  error: extractPublishIntentError(event.output),
+                }),
+              },
+              {
+                type: "setError",
+                params: ({ event }) => ({
+                  reason: "ERR_CANNOT_PUBLISH_INTENT",
+                  error: extractPublishIntentError(event.output),
```
cawabunga-bytes, line 530: You can actually skip setting error object as it does not provide valuable information. You always attach error specific data to this object:
```ts
// context
 error: null | {
    tag: "err"
    value: 
      {
        reason:
          | "ERR_USER_DIDNT_SIGN"
          // rest...
      } | { 
           reason: "ERR_CANNOT_PUBLISH_INTENT",
           server_reason: string 
      }
  }

// this action
                type: "setError",
                params: ({ event }) => ({
                  reason: "ERR_CANNOT_PUBLISH_INTENT",
                  server_reason: event.output.value,
                })
```

So you can get rid of `extractPublishIntentError` entirely.

src/features/machines/swapIntentMachine.ts
```diff
@@ -68,18 +67,23 @@ type Context = {
...
+            | "ERR_SIGNED_DIFFERENT_ACCOUNT"
+            | "ERR_PUBKEY_EXCEPTION"
+            | "ERR_CANNOT_PUBLISH_INTENT"
+            | "ERR_QUOTE_EXPIRED_RETURN_IS_LOWER"
+            | WalletErrorCode
+            | PublicKeyVerifierErrorCodes
+          error: Error | null
+        }
+      | {
+          reason: "ERR_CANNOT_PUBLISH_INTENT"
```
cawabunga-bytes, line 84: also need to remove `            | "ERR_CANNOT_PUBLISH_INTENT"`

depressedPlumber502: If I remove it from the value above there is a collision in swap form

PR #147, opened 2024-11-27
feat: add Dogecoin chain support

cawabunga-bytes:
Some thoughts about structure. It might better to have a special folder `chains`, which could contain chain specific functions. Now chain specific logic evenly spread in SDK.

Another point is address validation. Now we have regexp verification, which might not be great, since it may miss some rules. There are libraries like https://github.com/trezor/trezor-address-validator which solve this problem, but they all have some issues (e.g. no typescript, no solana, large size, etc). We may consider creating own library, maybe by forking that library and fixing all the issues.


jobotics:
Totally agree! Regarding Solana, I noticed this [commit](https://github.com/trezor/trezor-address-validator/commit/93e65f90ca1a7f4a0cca224f107516955d5d1e34), so it seems to be supported.

jobotics (approved):

PR #155, opened 2024-12-02
fix: token balance check

jobotics:
The goal of this PR is to improve code clarity by utilizing the appropriate token field and ensuring the correct balances are returned. Following by the [comments](https://github.com/defuse-protocol/defuse-sdk/pull/145).

### DIFF
--- a/src/features/deposit/components/DepositForm/index.tsx
+++ b/src/features/deposit/components/DepositForm/index.tsx
@@ -32,7 +32,11 @@ import {
   type SwappableToken,
 } from "../../../../types"
 import { formatTokenValue } from "../../../../utils/format"
-import { isBaseToken, isUnifiedToken } from "../../../../utils/token"
+import {
+  isBaseToken,
+  isNativeToken,
+  isUnifiedToken,
+} from "../../../../utils/token"
 import type { Context } from "../../../machines/depositUIMachine"
 import { DepositUIMachineContext } from "../DepositUIMachineProvider"
 import { Deposits } from "../Deposits"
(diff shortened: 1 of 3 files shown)


src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -678,17 +682,19 @@ function renderBalance(
...
 ) {
-  if (isBaseToken(token)) {
-    if (token.address === "wrap.near") {
-      return balance + nativeBalance
-    }
-    if (token.address === "native") {
-      return nativeBalance
-    }
-    return balance
+  // If the token is base and is wNEAR, return the combined balance of NEAR and wNEAR
```
cawabunga-bytes, line 685: Code should not explain what it does in most cases, because it just duplicates the code. Code may change and the comment most likely won't be changed, which will introduce confusion for a reader. It's better to explain why it does so.

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -678,17 +682,19 @@ function renderBalance(
...
+  if (isBaseToken(token) && token.address === "wrap.near") {
+    return balance + nativeBalance
   }
-  if (isUnifiedToken(token) && token.unifiedAssetId === "eth") {
+
+  // If the token is base and is native or it is unified and contains a native token, return the native balance
+  if (
+    (isBaseToken(token) && isNativeToken(token)) ||
+    (isUnifiedToken(token) &&
+      token.groupedTokens.some((t) => "type" in t && t.type === "native"))
```
cawabunga-bytes, line 694: You can simplify this `token.groupedTokens.some(isNative)`

cawabunga-bytes (approved):

jobotics:
@cawabunga-bytes Please don't merge this PR yet, I want to re-check Set Max value for Near + wNear

PR #156, opened 2024-12-03
chore: remove unnecesary tailwind dependecy

depressedPlumber502:
Screenshot 2024-12-03 at 12 14 10

Removes this from build logs


tailwind.config.js
```diff
@@ -1,5 +1,4 @@
 import forms from "@tailwindcss/forms"
-import lineClamp from "@tailwindcss/line-clamp"
```
cawabunga-bytes, line 2: You forgot to remove usage of it below.

depressedPlumber502: fixed

cawabunga-bytes (approved):

PR #160, opened 2024-12-05
feat: enabled deposit to turbochain

jobotics:
This update enables users to make direct deposits on Turbochain using EVM wallets such as MetaMask.

For the completion of this PR, the following updates are required:
- Add the `siloToSilo` contract to `silo.turbochain` at settings.

Example:
---
Screenshot 2024-12-05 at 19 26 09
Screenshot 2024-12-05 at 19 26 32

### DIFF
--- a/src/config/settings.ts
+++ b/src/config/settings.ts
@@ -36,7 +36,13 @@ export let settings: Settings = {
     solana:
       "https://quiet-solemn-asphalt.solana-mainnet.quiknode.pro/91aa5918df65520d47a54d88590f1503a4e804b1",
     dogecoin: "https://go.getblock.io/5f7f5fba970e4f7a907fcd2c5f4c38a2",
-    turbochain: "https://rpc-0x4e45415f.aurora-cloud.dev ",
+    turbochain: "https://rpc-0x4e45415f.aurora-cloud.dev",
+  },
+  /**
+   * SiloToSilo addresses on blockchains within Aurora Engine
+   */
+  silo: {
+    turbochain: "", // TODO: add address to deployed contract at turbochain
   },
 }
 
--- a/src/config/settings.ts
+++ b/src/config/settings.ts
@@ -38,12 +38,6 @@ export let settings: Settings = {
     dogecoin: "https://go.getblock.io/5f7f5fba970e4f7a907fcd2c5f4c38a2",
     turbochain: "https://rpc-0x4e45415f.aurora-cloud.dev",
   },
-  /**
-   * SiloToSilo addresses on blockchains within Aurora Engine
-   */
-  silo: {
-    turbochain: "", // TODO: add address to deployed contract at turbochain
-  },
 }
 
 export const getSettings = (): Settings => settings
--- a/src/features/deposit/components/DepositForm/index.tsx
+++ b/src/features/deposit/components/DepositForm/index.tsx
@@ -377,20 +377,16 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
             depositSolanaResult,
             depositTurboResult
           )}
-        {userAddress &&
-          (network === BlockchainEnum.NEAR ||
-            network === BlockchainEnum.TURBOCHAIN) && (
-            <Deposits
-              chainName={
-                network === BlockchainEnum.NEAR
-                  ? "near"
-                  : network === BlockchainEnum.TURBOCHAIN
-                    ? "turbochain"
-                    : null
-              }
-              depositResult={depositNearResult ?? depositTurboResult}
-            />
-          )}
+        <Deposits
+          chainName={
+            network === BlockchainEnum.NEAR
+              ? "near"
+              : network === BlockchainEnum.TURBOCHAIN
+                ? "turbochain"
+                : null
+          }
+          depositResult={depositNearResult ?? depositTurboResult}
+        />
         {network !== BlockchainEnum.NEAR && isDepositReceived && (
           <DepositSuccess />
         )}
--- a/src/features/deposit/components/DepositForm/index.tsx
+++ b/src/features/deposit/components/DepositForm/index.tsx
@@ -73,6 +73,7 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
   const depositNearResult = snapshot.context.depositNearResult
   const depositEVMResult = snapshot.context.depositEVMResult
   const depositSolanaResult = snapshot.context.depositSolanaResult
+  const depositTurboResult = snapshot.context.depositTurboResult
 
   const depositAddress =
     generatedAddressResult?.tag === "ok"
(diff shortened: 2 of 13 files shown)


src/config/settings.ts
```diff
@@ -36,7 +36,13 @@ export let settings: Settings = {
...
       "https://quiet-solemn-asphalt.solana-mainnet.quiknode.pro/91aa5918df65520d47a54d88590f1503a4e804b1",
     dogecoin: "https://go.getblock.io/5f7f5fba970e4f7a907fcd2c5f4c38a2",
-    turbochain: "https://rpc-0x4e45415f.aurora-cloud.dev ",
+    turbochain: "https://rpc-0x4e45415f.aurora-cloud.dev",
+  },
+  /**
+   * SiloToSilo addresses on blockchains within Aurora Engine
+   */
+  silo: {
+    turbochain: "", // TODO: add address to deployed contract at turbochain
```
cawabunga-bytes, line 45: I doubt about having a config for this, as it's not expected to be changed as it's bound to a chain. I would suggest moving it to just constants.

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -375,11 +374,23 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
...
+            <Deposits
+              chainName={
+                network === BlockchainEnum.NEAR
+                  ? "near"
+                  : network === BlockchainEnum.TURBOCHAIN
+                    ? "turbochain"
+                    : null
+              }
+              depositResult={depositNearResult ?? depositTurboResult}
+            />
```
cawabunga-bytes, line 392: This kind of confusing. So it renders result of deposit operation. I assume this will be a part of that refactoring along with machines.

Putting that aside I think checking `userAddress` and `network` is not needed here, you just need to render a whatever result the machine returned.

src/services/depositService.ts
```diff
@@ -141,6 +140,30 @@ export function createDepositEVMERC20Transaction(
   }
 }
 
+export function createDepositSiloToSiloTransaction(
```
cawabunga-bytes, line 143: nit: it should be `createDepositFromSiloTransaction`. Yes, it might use SiloToSilo contract, but it might not.

src/services/depositService.ts
```diff
@@ -141,6 +140,30 @@ export function createDepositEVMERC20Transaction(
...
+): { to: Address; data: Hash } {
+  // TODO check chain type for userAddress as a rule, pass through the user chainType
+  const data = encodeFunctionData({
+    abi: siloAbi,
+    functionName: "safeFtTransferCallToNear",
+    args: [
+      assetAccountId as Address,
+      amount,
+      depositAddress,
+      userAddressToDefuseUserId(userAddress, ChainType.EVM),
```
cawabunga-bytes, line 158: Please address these todo, it's important to not accidentally pass incorrect user id within the protocol.

src/services/depositService.ts
```diff
@@ -141,6 +140,30 @@ export function createDepositEVMERC20Transaction(
   }
 }
 
+export function createDepositSiloToSiloTransaction(
+  assetAccountId: string,
```
cawabunga-bytes, line 144: nit: it is `tokenAddress` / `assetAddress`, not account id

src/services/depositService.ts
```diff
@@ -141,6 +140,30 @@ export function createDepositEVMERC20Transaction(
...
+  amount: bigint,
+  depositAddress: string,
+  siloAddress: string
+): { to: Address; data: Hash } {
+  // TODO check chain type for userAddress as a rule, pass through the user chainType
+  const data = encodeFunctionData({
+    abi: siloAbi,
+    functionName: "safeFtTransferCallToNear",
+    args: [
+      assetAccountId as Address,
```
cawabunga-bytes, line 155: In order to have a proper type safety you can use `getAddress` from viem. It will also check validity of the address and throw if address is not correct.

src/features/machines/depositUIMachine.ts
```diff
@@ -255,6 +268,13 @@ export const depositUIMachine = setup({
         context.formValues.network === BlockchainEnum.SOLANA
       )
     },
+    isDepositTurbochainRelevant: ({ context }) => {
+      return (
+        context.balance + context.nativeBalance >=
```
cawabunga-bytes, line 273: I don't remember if I've asked you already, but why we add balance and native balance?

src/utils/chainTxExplorer.ts
```diff
@@ -0,0 +1,25 @@
...
+      return "https://explorer.turbo.aurora.dev/tx/"
+    case "bitcoin":
+      return "https://blockchain.info/tx/"
+    case "solana":
+      return "https://solscan.io/tx/"
+    case "dogecoin":
+      return "https://dogechain.info/tx/"
+    default:
+      blockchain satisfies never
+      return ""
```
cawabunga-bytes, line 23: Let's return null, as it reflect an empty value.

src/services/depositService.ts
```diff
@@ -362,3 +389,34 @@ export function getWalletRpcUrl(network: BlockchainEnum): string {
       throw new Error("exhaustive check failed")
   }
 }
+
+const siloAbi = [
```
cawabunga-bytes, line 393: It's siloToSiloABI

src/services/depositService.ts
```diff
@@ -283,11 +306,15 @@ export function getAvailableDepositRoutes(
...
       switch (network) {
         case BlockchainEnum.NEAR:
-        case BlockchainEnum.TURBOCHAIN:
           return {
             activeDeposit: false,
             passiveDeposit: false,
           }
+        case BlockchainEnum.TURBOCHAIN:
+          return {
+            activeDeposit: true,
```
cawabunga-bytes, line 315: So we'll have two types of active deposits. One is deposit from a silo, another is transfer using a bridge. Do you have plans how to implement it?

src/features/deposit/components/DepositForm/index.tsx
jobotics: Yes, you are right.

src/features/machines/depositUIMachine.ts
jobotics: This is only for the Near network, because if the user chooses Near for the deposit, then in this case his available balance is Near plus wNear. In this case it is a mistake, I will fix it.

jobotics: Also I'll do correction for `isDepositEVMRelevant` and `isDepositSolanaRelevant`

src/services/depositService.ts
jobotics: This can be just a generated address, just like a `passiveDeposit`, and the user will need to send the token there themselves, or we can add a variable, something as `bridgeDeposit`.

PR #163, opened 2024-12-06
fix: layout shift on swap input change

depressedPlumber502:
Fixes swap input layout shift on balance change

Removing some of the hardoced max width as it should be set by SDK consumer

### DIFF
--- a/src/components/SelectAssets.tsx
+++ b/src/components/SelectAssets.tsx
@@ -25,7 +25,7 @@ export const SelectAssets = ({ selected, handleSelect }: Props) => {
     <button
       type={"button"}
       onClick={handleAssetsSelect}
-      className="bg-white shadow-select-token rounded-full flex justify-between items-center p-1 gap-2.5 dark:bg-black-800 dark:shadow-select-token-dark"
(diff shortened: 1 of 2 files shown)


depressedPlumber502:
Have a feeling that all these errors fields are obsolete and I never seen theme

src/components/SelectAssets.tsx
```diff
@@ -25,7 +25,7 @@ export const SelectAssets = ({ selected, handleSelect }: Props) => {
     <button
       type={"button"}
       onClick={handleAssetsSelect}
-      className="max-w-[148px] md:max-w-[210px] bg-white shadow-select-token rounded-full flex justify-between items-center p-1 gap-2.5 dark:bg-black-800 dark:shadow-select-token-dark"
```
jobotics, line 28: We should take into account tokens with excessively long symbol names. Long text can stretch this element, disrupting the form layout and breaking the design.

Example:
Screenshot 2024-12-09 at 10 50 08

depressedPlumber502: Returned the fixed size for token selector and the issue is gone

jobotics (approved):

PR #165, opened 2024-12-09
feat: w-full widgets

depressedPlumber502:


depressedPlumber502:
> Please add some reasonable max size.
And what if eventually someone wants to have a bigger size than the one set by me? What do they do? Open a PR for passing className string to override the value ?

cawabunga-bytes:
I doubt consumers will want to have 1000px widget, what do you think?

depressedPlumber502:
Set default width to 472px as before

PR #167, opened 2024-12-10
fix: eth to near withdraw storage_deposit error

depressedPlumber502:
ETH on Near doesn't require storage deposit

### DIFF
--- a/src/services/withdrawService.ts
+++ b/src/services/withdrawService.ts
@@ -199,6 +199,14 @@ async function determineNEP141StorageRequirement(
     return { tag: "ok", value: null }
   }
 
+  // For withdrawing ETH to NEAR no storage_deposit is required. (P.S. aurora is ETH address on Near network)
+  if (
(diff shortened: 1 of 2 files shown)


src/services/withdrawService.ts
```diff
@@ -199,6 +199,14 @@ async function determineNEP141StorageRequirement(
     return { tag: "ok", value: null }
   }
 
+  // For withdrawing ETH to NEAR no storage_deposit is required. (P.S. aurora is ETH address on Near network)
+  if (
+    formValues.tokenOut.chainName === "near" &&
+    formValues.tokenOut.address === "aurora"
+  ) {
+    return { tag: "ok", value: null }
+  }
```
cawabunga-bytes, line 208: I think this check should be moved to `getNEP141StorageRequired`, since that is general function for getting storage. This functions handles withdrawal specific logic for storage.

cawabunga-bytes (approved):

PR #174, opened 2024-12-11
chore: update creation deposit from silo

jobotics:


src/constants/tokens.ts
```diff
@@ -1,5 +1,7 @@
 import type { BaseTokenInfo } from "../types/base"
 
+export const NATIVE_TOKEN_ADDRESS = "0x0000000000000000000000000000000000000000"
```
cawabunga-bytes, line 3: It's exclusively for AuroraEngine SiloToSilo contract, I suggest putting in place where it's used.

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -240,12 +240,14 @@ export function DepositUIMachineProvider({
                 )
 
                 const tx = createDepositFromSiloTransaction(
-                  tokenAddress,
+                  tokenAddress === "native"
```
cawabunga-bytes, line 243: `tokenAddress` won't be `native` in the future.

cawabunga-bytes (approved):
`tokenAddress`

cawabunga-bytes:
Fix types please

src/features/deposit/components/DepositUIMachineProvider.tsx
jobotics: noted

PR #179, opened 2024-12-12
chore: add check for allowance and approve transaction for erc20

jobotics:


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -239,6 +241,26 @@ export function DepositUIMachineProvider({
...
                 )
 
+                if (tokenAddress !== "native") {
+                  const allowance = await getAllowance(
+                    tokenAddress,
+                    accountId,
+                    siloToSiloAddress.turbochain,
+                    BlockchainEnum.TURBOCHAIN
+                  )
+                  assert(allowance != null, "Allowance is not defined")
```
cawabunga-bytes, line 251: You'll log the same error twice when allowance request fails: in the caller of this function and getAllowance itself.

src/services/depositService.ts
```diff
@@ -271,6 +273,45 @@ export async function getMinimumStorageBalance(
...
 }
 
+export async function getAllowance(
+  tokenAddress: string,
+  owner: string,
+  spender: string,
+  network: BlockchainEnum
+): Promise<bigint | null> {
+  try {
+    const client = createPublicClient({
```
cawabunga-bytes, line 283: I don't like creating an rpc client in place, since it may send unnecessary request every time it is created. For example, ethers provider sends eth_chainId request for the first call. It's place to think of and generalize the approach so it can be used in other place as well.

cawabunga-bytes (approved):

PR #184, opened 2024-12-12
feat: publish intent retry

depressedPlumber502:
Adding retires for intent publishing.

This library is used - https://github.com/lifeomic/attempt
It has 0 deps

Some context for params usage - https://encore.dev/blog/retries

Some example of timing I get for retries

Screenshot 2024-12-12 at 18 39 08

a small code sample to try it locally

```
import { retry } from "@lifeomic/attempt"

let counter = 0
async function main() {
  await retry(
    async () => {
      const a = new Date()
      console.log(
        counter,
        `${a.getHours()}:${a.getMinutes()}:${a.getSeconds()}:${a.getMilliseconds()}`
      )

      if (counter > 123) {
        return
      }
      counter++
      throw new Error()
    },
    {
      delay: 1000,
      factor: 1.5,
      maxAttempts: 7,
      jitter: true,
      minDelay: 1000,
    }
  )
}

main()
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
     "uuid": "^10.0.0",
     "viem": "^2.18.1",
     "xstate": "^5.18.2",
-    "zustand": "^5.0.0-rc.2"
+    "zustand": "^5.0.0-rc.2",
+    "@lifeomic/attempt" : "3.1.0"
   },
   "devDependencies": {
     "@biomejs/biome": "^1.9.3",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "prepare": "husky && ./scripts/gen-defuse-types.sh"
   },
   "dependencies": {
+    "@lifeomic/attempt": "^3.1.0",
     "@noble/curves": "1.4.0",
     "@radix-ui/react-checkbox": "^1.1.1",
     "@radix-ui/react-dialog": "1.1.1",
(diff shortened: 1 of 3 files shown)


src/services/intentService.ts
```diff
@@ -16,15 +17,26 @@ export async function publishIntent(
...
+      return { tag: "err", value: { reason: result.status } }
+    },
+    {
+      delay: 1000,
+      factor: 1.5,
+      maxAttempts: 7,
+      jitter: true,
+      minDelay: 1000,
+    }
+  )
```
cawabunga-bytes, line 39: The order should be slightly different. Handling should go after retry.
```js
const res = await retry(() => myFn())
doSomething(res)
```

package.json
```diff
@@ -55,7 +55,8 @@
     "uuid": "^10.0.0",
     "viem": "^2.18.1",
     "xstate": "^5.18.2",
-    "zustand": "^5.0.0-rc.2"
+    "zustand": "^5.0.0-rc.2",
+    "@lifeomic/attempt" : "3.1.0"
```
cawabunga-bytes, line 59: Please reinstall using `yarn add @lifeomic/attempt`, it will write correct version and put in right order

src/services/intentService.ts
depressedPlumber502: But it worked hm

depressedPlumber502:
Comments fixed but it is weird but using dev shell script the FE can't find the module

@cawabunga-bytes

cawabunga-bytes (approved):

PR #202, opened 2024-12-17
refactor: nep141StorageService

depressedPlumber502:
Refactors the function that gets the necessary amount of near user need to pay for storage deposit

P.S. DO NOT MERGE STRAIGHTAWAY IF OK


src/services/nep141StorageService.ts
```diff
@@ -17,33 +15,30 @@ export type Output =
...
 export async function getNEP141StorageRequired({
-  token,
+  tokenAddress,
   userAccountId,
 }: {
-  token: BaseTokenInfo
+  tokenAddress: string
   userAccountId: string
 }): Promise<Output> {
-  if (token.chainName !== "near" || isNativeToken(token)) {
```
depressedPlumber502, line 31: I believe we should not have this in this function. It makes it is hard to reason about it and introduces unnecessary complexity, which can be avoided easily.

src/services/withdrawService.ts
```diff
@@ -347,21 +346,25 @@ async function getBalances(
...
+  formValues: WithdrawFormContext
+}): Promise<
   | { tag: "ok"; value: bigint }
   | { tag: "err"; value: { reason: "ERR_NEP141_STORAGE" } }
 > {
   assert(formValues.parsedRecipient != null, "parsedRecipient is null")
 
+  if (
+    formValues.tokenOut.chainName !== "near" ||
+    isNativeToken(formValues.tokenOut)
```
depressedPlumber502, line 361: What edge case is covered by isNativeToken function here ?
@cawabunga-bytes

src/services/nep141StorageService.ts
cawabunga-bytes: So you just moved this logic somewhere else, you didn't reduce complexity.

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -105,34 +104,31 @@ export function DepositUIMachineProvider({
 
                 let tx: Transaction["NEAR"][] = []
 
+                const storageDepositPayment = await getNEP141StorageRequired({
+                  tokenAddress,
```
cawabunga-bytes, line 108: Logic above (extracting address of the token) is not fully correct. It does not handle a case when token is native, and technically it doesn't have an address. You need to consider it as well.

src/services/nep141StorageService.ts
depressedPlumber502: the purpose for this function in my opinion should be to deal with only nep141 and nothing else, why the function should be aware of any other chains beside NEAR ?

depressedPlumber502: Having this function reason about diff chains deff not making the whole thing simpler

depressedPlumber502: To have this function accepting token like before it would also force to make us have all sorts of weird checks to handle the typing

depressedPlumber502: I would rather have a different function all together that accepts an abstract token and says if storage_deposit is even needed

src/features/deposit/components/DepositUIMachineProvider.tsx
depressedPlumber502: As far as I see it only deals with NEAR tokens, and when Native near is accepted in our code we work with it as wrap.near .

PLS correct me if I am wrong

src/services/withdrawService.ts
depressedPlumber502: Screenshot 2024-12-19 at 13 02 21


Can this be handled like this ?

depressedPlumber502: I would rather have a separate function that accepts a token and return a bool that says if storage deposit is necessary for this token

depressedPlumber502: And put there all the edge cases and checks

src/features/deposit/components/DepositUIMachineProvider.tsx
cawabunga-bytes: FungibleTokenInfo and NativeTokenInfo both have `address` property (for some specific reasons). In case of native token address is "native". 
> when Native near is accepted in our code we work with it as wrap.near .
Not correct, for example, during deposit we accept just NEAR

src/services/nep141StorageService.ts
cawabunga-bytes: So you just want to create another function with the same logic. Current function is not that big to have concerns about it.

src/services/nep141StorageService.ts
```diff
@@ -17,7 +17,8 @@ export type Output =
 
 /**
  * Get the amount of NEP-141 storage required for the user to store the token.
- * @param token The token to check.
+ * @param tokenAddress.
+ * @param tokenChainName.
```
cawabunga-bytes, line 21: Please revert this.

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -95,44 +94,38 @@ export function DepositUIMachineProvider({
...
-                  : (asset.groupedTokens.find(
+                const tokenToDeposit = isBaseToken(asset)
+                  ? asset
+                  : asset.groupedTokens.find(
                       (token) => token.chainName === "near"
-                    )?.address ?? null)
+                    )
 
-                assert(tokenAddress != null, "Token address is not defined")
+                assert(tokenToDeposit, "Token address is not defined")
```
cawabunga-bytes, line 103: Can you update message here?

src/services/nep141StorageService.ts
lampamazaza: Fixed

depressedPlumber502: Fixed

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -95,44 +94,38 @@ export function DepositUIMachineProvider({
...
+
+                if (tokenToDeposit.address === "wrap.near") {
                   tx = createBatchDepositNearNativeTransaction(
-                    tokenAddress,
                     amount,
-                    amount - balance,
-                    isWrapNearRequired,
-                    minStorageBalance
+                    // If user does not have enough wrap.near we calculate how much native NEAR we need to wrap to match the amount to deposit
+                    amount - (balance || 0n),
```
cawabunga-bytes, line 121: If balance is larger than amount, then it will be negative value

cawabunga-bytes (approved):
Nice work!

PR #203, opened 2024-12-17
feat: Add Aurora chain support

cawabunga-bytes:


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -268,25 +268,28 @@ export function DepositUIMachineProvider({
                 } = input
 
                 const chainId = getEVMChainId(chainName)
+                const siloToSiloAddress_ =
```
jobotics, line 271: `_` is kind of private sign for variable, when is used at front. Is it strong reason to use it or we can rename it to just into camel case?

jobotics (approved):

PR #212, opened 2024-12-19
fix: add error when deposit amount less than min allowed

jobotics:
Relevant to Solana only.

### DIFF
--- a/src/features/deposit/components/DepositUIMachineProvider.tsx
+++ b/src/features/deposit/components/DepositUIMachineProvider.tsx
@@ -227,9 +227,13 @@ export function DepositUIMachineProvider({
           depositSolanaActor: depositSolanaMachine.provide({
             actors: {
               signAndSendTransactions: fromPromise(async ({ input }) => {
-                const { amount, depositAddress, accountId } = input
+                const { amount, depositAddress, accountId, asset } = input
 
                 assert(depositAddress != null, "Deposit address is not defined")
+
+                if (isBaseToken(asset)) {
+                  assert(amount >= BigInt(12500000), "Minimum amount is 0.0125")
+                }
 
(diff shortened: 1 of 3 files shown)


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -227,10 +227,14 @@ export function DepositUIMachineProvider({
           depositSolanaActor: depositSolanaMachine.provide({
             actors: {
               signAndSendTransactions: fromPromise(async ({ input }) => {
-                const { amount, depositAddress, accountId } = input
+                const { amount, depositAddress, accountId, asset } = input
 
                 assert(depositAddress != null, "Deposit address is not defined")
 
+                if (isBaseToken(asset)) {
+                  assert(amount >= BigInt(12500000), "Minimum amount is 0.0125")
```
cawabunga-bytes, line 235: I don't think it's a good place for validating the data, here you're just sending a transaction. Validation should be done prior that. Also we need to check minimum value in the form too and minimal amount should be taken from the bridge api (if the bridge involved).

jobotics: Right, but at the same time, we don't have the right place for it, meaning we have, let's say root deposit machine  `depositUIMachine` that isn't a good fit for it and we have `depositSolanaMachine` which is relevant but we plan to refactor it and which isn't ready to mutate with this validation now. 

I have some thinking about that, all our deposit action machines should be rewritten to one single generic `depositBroadcastMachine` which should work only for action event execution and validation transactions and that's it (without additional async requests, like stogareDeposit and approve, .etc). With this, we can shift the preparation work one step back before the user action event, and this will cut corners with has known issue with wallet trust execution event.

cawabunga-bytes (approved):

PR #220, opened 2025-01-06
fix: deposit NEAR when user never had wNEAR

depressedPlumber502:
A quick PR for fixing a NEAR deposit with 0 storageDeposit paid for wNEAR

### DIFF
--- a/src/services/depositService.ts
+++ b/src/services/depositService.ts
@@ -98,7 +98,7 @@ export function createBatchDepositNearNativeTransaction(
 ): Transaction["NEAR"][] {
   const actions: Transaction["NEAR"]["actions"] = []
 
-  if (nearAmountToWrap > 0n) {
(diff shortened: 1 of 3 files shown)


src/services/depositService.ts
```diff
@@ -98,7 +98,7 @@ export function createBatchDepositNearNativeTransaction(
 ): Transaction["NEAR"][] {
   const actions: Transaction["NEAR"]["actions"] = []
 
-  if (nearAmountToWrap > 0n) {
+  if (nearAmountToWrap > 0n || storagePayment > 0n) {
```
cawabunga-bytes, line 101: I remember that `storagePayment` could be negative. Could you double check it?

depressedPlumber502: Even if it is negative it won't pass this condition

depressedPlumber502: I added a check outside this function

PR #222, opened 2025-01-06
fix: deposit near with no wNear at all

depressedPlumber502:


cawabunga-bytes (approved):
Overall negative values should have not been returned in the first place.

jobotics (approved):

PR #228, opened 2025-01-08
feat: add usd prices to token fields

depressedPlumber502:
A demo for review of how I can add USD prices to input fields.

A few things left to add 

1. Formatting for really big USD amounts to avoid scientific notation in UI. Probably using Intl or some other way like 10mil 223th.
2. I ll add overall formatting to avoid a lot of digits after the decimal point.

Screenshot 2025-01-08 at 15 42 35
Screenshot 2025-01-08 at 15 42 40
Screenshot 2025-01-08 at 15 43 35

### DIFF
--- a/src/components/Form/FieldComboInput.tsx
+++ b/src/components/Form/FieldComboInput.tsx
@@ -102,7 +102,7 @@ export const FieldComboInput = <T extends FieldValues>({
   return (
     <div
       className={clsx(
-        "relative flex flex-col px-5 pt-5 pb-10 w-full bg-gray-50 dark:bg-black-900 dark:border-black-950",
+        "relative flex flex-col px-5 pt-5 pb-6 w-full bg-gray-50 dark:bg-black-900 dark:border-black-950",
         className
       )}
     >
@@ -119,7 +119,7 @@ export const FieldComboInput = <T extends FieldValues>({
           disabled={disabled}
           autoComplete="off"
           className={clsx(
-            "bg-gray-50 w-full text-3xl font-medium placeholder-black border-transparent focus:border-transparent focus:ring-0 dark:bg-black-900 dark:placeholder-white",
+            "bg-gray-50 w-full text-3xl font-medium placeholder-black border-transparent focus:border-transparent focus:ring-0 dark:bg-black-900 dark:placeholder-white px-0",
             disabled &&
               "text-black-200 pointer-events-none placeholder-black-200",
             {
@@ -133,23 +133,26 @@ export const FieldComboInput = <T extends FieldValues>({
         )}
       </div>
 
-      {fieldError ? (
-        <span className="absolute bottom-4 left-5 text-xs sm:text-sm font-medium text-red-400">
-          {(fieldError as FieldError).message}
-        </span>
-      ) : usdAmount ? (
-        <span className="absolute bottom-4 left-5 text-xs sm:text-sm font-medium text-gray-400">
-          {usdAmount}
-        </span>
-      ) : null}
-      {balance != null && (
-        <BlockMultiBalances
-          balance={balance}
-          decimals={selected?.decimals ?? 0}
-          handleClick={handleSetMaxValue}
-          disabled={disabled}
-        />
-      )}
+      <div className="flex justify-between items-center gap-2 mt-1 min-h-5 w-full max-w-[calc(100vw-106px)]">
+        {fieldError ? (
+          <span className="text-xs sm:text-sm font-medium text-red-400  whitespace-nowrap overflow-hidden">
+            {(fieldError as FieldError).message}
+          </span>
+        ) : usdAmount ? (
+          <span className="text-xs sm:text-sm font-medium text-gray-400 whitespace-nowrap overflow-hidden">
+            {usdAmount}
+          </span>
+        ) : null}
+        {balance != null && (
+          <BlockMultiBalances
+            balance={balance}
+            decimals={selected?.decimals ?? 0}
+            handleClick={handleSetMaxValue}
+            disabled={disabled}
+            className="ml-auto"
+          />
+        )}
+      </div>
     </div>
   )
 }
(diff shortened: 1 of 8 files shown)


src/components/Form/FieldComboInput.tsx
```diff
@@ -131,9 +133,13 @@ export const FieldComboInput = <T extends FieldValues>({
...
       </div>
 
-      {errors?.[fieldName] ? (
+      {fieldError ? (
         <span className="absolute bottom-4 left-5 text-xs sm:text-sm font-medium text-red-400">
-          {(errors[fieldName] as FieldError).message}
+          {(fieldError as FieldError).message}
+        </span>
+      ) : usdAmount ? (
+        <span className="absolute bottom-4 left-5 text-xs sm:text-sm font-medium text-gray-400">
```
depressedPlumber502, line 141: The color here is approximate not an exact value from figma. The figma one comes from radix color set , not sure how to get it out of there or should I just use there Text component or just set a hardcode value for now to avoid using radix at all ?

src/utils/getTokenUsdPrice.ts
```diff
@@ -0,0 +1,34 @@
+import type { TokenUsdPriceData } from "src/hooks/useTokensUsdPrices"
+import type { SwappableToken } from "src/types/swap"
+import { isBaseToken, isUnifiedToken } from "./token"
+
+const getTokenUsdPrice = (
```
depressedPlumber502, line 5: I am really not sure about this function or where to put it. But whatever way we decide to use prices we ll have to differentiate base token and unified token.

Maybe we should just separate token extraction and multipication into separate ones

src/utils/getTokenUsdPrice.ts
```diff
@@ -0,0 +1,34 @@
...
+    !tokenAmount ||
+    Number.isNaN(+tokenAmount)
+  )
+    return null
+  let tokenUsdPriceData = null
+  if (isBaseToken(token) && tokensUsdPriceData[token.defuseAssetId]) {
+    tokenUsdPriceData = tokensUsdPriceData[token.defuseAssetId]
+  } else if (isUnifiedToken(token)) {
+    for (const groupedToken of token.groupedTokens) {
+      if (
```
depressedPlumber502, line 22: I just pick the first one that matches in case of a Unified token (Swap)

depressedPlumber502:
Getting back to that discussion of Should we use a hook ?

Maybe we can just put a storage of these prices inside a machine and insert a fetching logic implementation as an argument. (In our case it can be raw query function from tan stack to also include caching there)

The utility function to extract the token price can be kept alongside the machine , although I reall not sure about it

depressedPlumber502:
@cawabunga-bytes pls let me now what you think
P.S. I know you not gonna love some of those things especially the utility function :D

jobotics:
We don't need so many decimals, as this represents a dollar equivalent, and values less than 1 cent are negligible. Therefore, the second digit after the decimal point is sufficient.

depressedPlumber502:
> We don't need so many decimals, as this represents a dollar equivalent, and values less than 1 cent are negligible. Therefore, the second digit after the decimal point is sufficient.

Yeah I mentioned that in the description. I am more concerned about whether it is better to put it the machine after all

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -16,10 +16,12 @@ import { QRCodeSVG } from "qrcode.react"
 import { useEffect, useState } from "react"
 import CopyToClipboard from "react-copy-to-clipboard"
 import { Controller, useFormContext } from "react-hook-form"
+import { useTokensUsdPrices } from "src/hooks/useTokensUsdPrices"
 import {
   assetNetworkAdapter,
   reverseAssetNetworkAdapter,
 } from "src/utils/adapters"
+import getTokenUsdPrice from "src/utils/getTokenUsdPrice"
```
jobotics, line 25: A bit confusing for me that we have hook `useTokensUsdPrices` and utility function `getTokenUsdPrice`, which lexically should do same work.

src/utils/getTokenUsdPrice.ts
jobotics: We could use the derived tokens, as we do for deposits and withdrawals, but since there’s no equivalent in swaps due to the absence of network usage, I think extraction is fine

src/features/deposit/components/DepositForm/index.tsx
depressedPlumber502: I just need some place to put this utility thing to get the token price out of the hook data

src/utils/getTokenUsdPrice.ts
```diff
@@ -0,0 +1,38 @@
...
+  tokenAmount: string,
+  token: SwappableToken | null,
+  tokensUsdPriceData?: TokenUsdPriceData
+): number | null => {
+  try {
+    if (
+      !tokensUsdPriceData ||
+      !token ||
+      !tokenAmount ||
+      Number.isNaN(+tokenAmount)
```
jobotics, line 15: Here we don't handle invalid inputs like Infinity or -Infinity. If you put on swap tokenIn 1.8e+308 result would be something ~$∞, which doesn't make sense to the user

src/features/deposit/components/DepositForm/index.tsx
depressedPlumber502: If you can come up with a better name I would be glad to use it +)

src/features/swap/components/SwapForm.tsx
```diff
@@ -191,6 +207,7 @@ export const SwapForm = ({ onNavigateDeposit }: SwapFormProps) => {
           errors={errors}
           disabled={true}
           isLoading={snapshot.matches({ editing: "waiting_quote" })}
+          usdAmount={usdAmountOut ? `~${formatUsdAmount(usdAmountOut)}` : ""}
```
jobotics, line 210: If we don't have `usdAmountOut` then I think it should be `null` instead of empty `string`

depressedPlumber502: fixed

jobotics (approved):

PR #232, opened 2025-01-10
fix: remove deprecated field from native token info

jobotics:
Remove `address` from `NativeTokenInfo` interface


src/features/machines/depositEstimationActor.ts
```diff
@@ -155,9 +155,13 @@ export const depositEstimationMachine = setup({
...
         src: "estimateMaxDepositValueActor",
         input: ({ event }) => {
+          const address = isFungibleToken(event.params.token)
+            ? event.params.token.address
+            : null
+          assert(address != null, "Address is not defined")
           return {
             ...event.params,
-            tokenAddress: event.params.token.address,
+            tokenAddress: address,
```
cawabunga-bytes, line 164: Looks like `tokenAddress` is not unnecessary, because you already pass `token` as input. It might be better just to change the input from BaseTokenInfo to FungibleTokenInfo.

src/features/machines/getBalanceMachine.ts
```diff
@@ -112,12 +112,15 @@ function mapTokenList(
...
 ): Array<[string, string]> {
   return tokenList.reduce<Array<[string, string]>>((acc, token) => {
-    if (isBaseToken(token)) {
+    if (isFungibleToken(token)) {
       acc.push([token.defuseAssetId, token.address])
     }
     if (isUnifiedToken(token)) {
       for (const groupToken of token.groupedTokens) {
-        acc.push([groupToken.defuseAssetId, groupToken.address])
+        // As this map for nep141 tokens, we can map the fungible tokens only
```
cawabunga-bytes, line 120: I didn't get this, could you explain?

cawabunga-bytes (commented):
Please review my comments, they are worth to addressing.

src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -108,15 +108,19 @@ export function DepositUIMachineProvider({
...
+                assert(address != null, "Address is not defined")
                 assert(
                   storageDepositRequired !== null,
                   "Storage deposit required is null"
                 )
                 assert(nearBalance !== null, "Near balance is null")
 
                 let tx: Transaction["NEAR"][] = []
-                if (derivedToken.address === "wrap.near") {
+                if (address === "wrap.near") {
```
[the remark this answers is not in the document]
jobotics: In what case might we need this? This is a specific case for wrap.near only, so we shouldn't handle any other native tokens here. If we do, it indicates an issue with the NEAR deposit logic. In 99% of cases, we don't need to calculate how much NEAR native balance to wrap.

src/features/machines/depositEstimationActor.ts
jobotics: Yes, you're right. I saw this, but it requires some refactoring, which I’d prefer not to do under this ticket.

src/features/machines/getBalanceMachine.ts
jobotics: Yes, `groupedTokens` here could be either `FungibleTokenInfo` of `NativeTokenInfo`, and we use this map to get nep141 balance only, like that:
```           
[tokenId]: await getNearNep141Balance({
              tokenAddress,
              accountId,
            }),
```
so there's no point in doing this with the native token

cawabunga-bytes (approved):

PR #235, opened 2025-01-14
feat: handle insufficient amount

depressedPlumber502:
A draft of  adding insufficient amount   returned in format `{ type: "INSUFFICIENT_AMOUNT" }` from solver


src/features/machines/swapIntentMachine.ts
```diff
@@ -334,10 +337,11 @@ export const swapIntentMachine = setup({
 
   on: {
     NEW_QUOTE: {
+      guard: "isSuccesfulQuote",
       actions: [
         {
           type: "proposeQuote",
-          params: ({ event }) => event.params.quote,
+          params: ({ event }) => event.params.quote.value as AggregatedQuote,
```
jobotics, line 344: We should avoid any type casting. Probably the NEW_QUOTE event shouldn't be expected to have a non-AggregatedQuote type.
Or you can use assert before, something like this:
```
assert(event.params.quote.tag === "ok")
```

src/services/withdrawService.ts
```diff
@@ -195,7 +197,8 @@ async function determineNEP141StorageRequirement(
       value: {
         reason:
           | "ERR_NEP141_STORAGE"
-          | "ERR_EMPTY_QUOTE"
+          | FailedQuote["type"]
```
jobotics, line 200: This seems incorrect. Why not just specify the error type as INSUFFICIENT_AMOUNT? Also, since it is still an error, it should start with ERR_.

depressedPlumber502: I want to use same Text value as from solver and not mapping one to another

src/features/machines/swapIntentMachine.ts
depressedPlumber502: Feels just like a workaround around guard not adjusting the type.

Assert would also require an onError handler

jobotics: Yes, kind of. As I mentioned, you probably need to adjust the NEW_QUOTE event so that it is triggered only when an aggregate quote is available. If this change is made, the guard would no longer be necessary

src/services/quoteService.ts
```diff
@@ -3,7 +3,17 @@ import { logger } from "../logger"
...
 import { computeTotalBalance } from "../utils/tokenUtils"
 import { quote } from "./solverRelayHttpClient"
-import type { QuoteResponse } from "./solverRelayHttpClient/types"
+import type {
+  FailedQuote,
+  Quote,
+  QuoteResponse,
+} from "./solverRelayHttpClient/types"
+
+export function isFailedQuote(
```
cawabunga-bytes, line 12: nit: no export needed

src/services/quoteService.ts
```diff
@@ -27,24 +37,27 @@ export interface AggregatedQuote {
...
-  quoteHashes: [],
-  expirationTime: new Date(0).toISOString(),
-  totalAmountIn: 0n,
-  totalAmountOut: 0n,
-  amountsIn: {},
-  amountsOut: {},
-  tokenDeltas: [],
-})
-
+export type QuoteRequestResult =
```
cawabunga-bytes, line 40: nit: maybe just quoteResult?

src/services/quoteService.ts
```diff
@@ -110,30 +133,63 @@ export async function queryQuoteExactOut(
...
+        tokenDeltas: [
+          [input.tokenIn, -BigInt(bestQuote.amount_in)],
+          [input.tokenOut, BigInt(bestQuote.amount_out)],
+        ],
+      },
+    }
+  }
+
+  if (failedQuotes[0]) {
+    return {
```
cawabunga-bytes, line 182: nit: I think it's possible to return early, when you first met failed quote. Am I missing something?

src/services/solverRelayHttpClient/types.ts
```diff
@@ -30,15 +30,25 @@ export type QuoteRequest = JSONRPCRequest<
 
 export type Params<T extends JSONRPCRequest<unknown, unknown>> = T["params"][0]
 
-export type QuoteResponse = JSONRPCResponse<null | Array<{
+export enum FAILED_QUOTES_TYPES {
+  INSUFFICIENT_AMOUNT = "INSUFFICIENT_AMOUNT",
+}
+export type InsufficientAmountQuote = {
+  type: FAILED_QUOTES_TYPES.INSUFFICIENT_AMOUNT
+  min_amount: number
```
cawabunga-bytes, line 38: Double check the `number` type please

src/services/withdrawService.ts
```diff
@@ -148,8 +149,9 @@ export async function prepareWithdraw(
     { signal }
   )
 
+  assert(swapRequirement?.swapQuote.tag === "ok", "Fail")
   const receivedAmount = calcWithdrawAmount(
-    swapRequirement?.swapQuote ?? null,
+    swapRequirement?.swapQuote.value ?? null,
```
cawabunga-bytes, line 154: nit: null checks are not need here (if I'm not mistaken)

src/services/withdrawService.ts
```diff
@@ -54,7 +51,8 @@ export type PreparationOutput =
               | "ERR_NEP141_STORAGE"
               | "ERR_CANNOT_FETCH_POA_BRIDGE_INFO"
               | "ERR_CANNOT_FETCH_QUOTE"
-              | "ERR_EMPTY_QUOTE"
+              | "NO_QUOTES"
+              | "INSUFFICIENT_AMOUNT"
```
cawabunga-bytes, line 55: nit: general thoughts, no changes needed. Error reasons need more specification, right now it's hard to understand where error came from. For example, ERR_{FEATURE}_{ENTITY} might be a better form: `ERR_QUOTE_NO_QUOTE`, `ERR_QUOTE_CANT_FETCH` etc.

src/features/swap/components/SwapUIMachineProvider.tsx
```diff
@@ -83,13 +82,13 @@ export function SwapUIMachineProvider({
             const quote = context.quote
             if (quote == null) {
               resetField("amountOut")
-            } else if (isAggregatedQuoteEmpty(quote)) {
+            } else if (quote.tag === "err") {
               setValue("amountOut", "–", {
                 shouldValidate: false,
               })
-            } else {
+            } else if (quote.tag === "ok") {
```
cawabunga-bytes, line 89: nit: condition is not needed here, because of sum-type.

src/services/withdrawService.ts
jobotics: Why do we need to stick to them?

cawabunga-bytes (commented):
Great job! In general this is valid approach. Some minor tweaks are needed, but I think already know those points.

src/services/withdrawService.ts
depressedPlumber502: In order to not create unnecessary mapping between solver errors and FE errors

src/services/quoteService.ts
depressedPlumber502: What if there is one failed quote and one valid one. I thought in case there is at least one valid it is better to return it?

src/features/machines/swapIntentMachine.ts
depressedPlumber502: Overall it won't be sent to from the Withdraw UI machine or Swap UI Machine, because I added a guard there. 

Maybe I should create a separate type for intent machine to explain that it only accepts  Valid ones, but it still won't protect us from sending smth invalid there  and it makes sense to have a guard

depressedPlumber502: @cawabunga-bytes what do you think ?

cawabunga-bytes: Overall typecasting is the last resort, otherwise just use assert (no huge performance overhead, easy to read), but help all the time when your assumption about the data is not correct.

depressedPlumber502: Can I just add a check with IF statement as assert throws an Error?

cawabunga-bytes: Why don't you like `assert`?

depressedPlumber502: I ll have to add an onError there otherwise a UI is getting crashed

PR #242, opened 2025-01-17
feat: hold intent if optimistic flow

jobotics:
This PR enables an optimistic flow by holding the signed data and using it later in the waiting pool. The data is revealed and broadcasted at the appropriate time.

Important: If the feature is disabled, there will be no changes to the previous flow.

<details>

<summary>Machine changes:</summary>

Screenshot 2025-01-17 at 14 06 45

</details>

### DIFF
--- a/src/features/machines/swapIntentMachine.ts
+++ b/src/features/machines/swapIntentMachine.ts
@@ -47,7 +47,7 @@ export type NEP141StorageRequirement =
       quote: null
     }
 
-type IntentOperationParams =
+export type IntentOperationParams =
   | {
       type: "swap"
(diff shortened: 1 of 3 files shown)


depressedPlumber502:
Is this for showing this inflight balances ?

jobotics:
> Is this for showing these inflight balances ? If yes then it feels like we over-complicate the machines just to have this balance thing rather then redoing balances themselves.

No, this is for holding the intent in the waiting pool until the balance becomes sufficient to execute it, and then executing it using the same flow as before.

I'm trying to minimize complexity as much as possible while keeping the same flow.

PR #248, opened 2025-01-21
feat: change release flow

depressedPlumber502:
**Goals**

1. Block any MAJOR release
2. Push to NPM registry instead of GITHUB REGISTRY 
3. Drop version to 0.0.1 as there won't be any MAJOR release any time soon (or 1.0.0?)

### DIFF
--- a/commitlint.config.cjs
+++ b/commitlint.config.cjs
@@ -1,10 +1,6 @@
 /** @type {import('@commitlint/types').UserConfig} */
 const config = {
   extends: ["@commitlint/config-conventional"],
-  rules: {
-    "body-max-line-length": [0],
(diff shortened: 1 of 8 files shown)


commitlint.config.cjs
```diff
@@ -1,10 +1,6 @@
 /** @type {import('@commitlint/types').UserConfig} */
 const config = {
   extends: ["@commitlint/config-conventional"],
-  rules: {
-    "body-max-line-length": [0],
-    "footer-max-line-length": [0],
-  },
```
cawabunga-bytes, line 7: This was added by purpose. Otherwise "squash and merge" won't trigger release, because the release action will yell that commit message is not correct.

PR #274, opened 2025-01-31
feat: add basic slippage functionality

cawabunga-bytes:
We have 2 pieces of functionailty that are affected by slippage. Swaps and withdraws (because  80% of withdraws are swap + withdraw).

## Swaps

No caveats in swaps. Users just sign a message with a little worse received amount. If there is a better quote, then the better quote will be used. Solver relay is responsible for transferring remaining amount.

## Withdraws

Here are 2 tricky things.

### 1. Withdraws of natively multichain assets. 

Swaps between natively multichain assets are 1-1. So there should not be any slippage. Eventually there will but now implementation assumes there aren't.

### 2. Withdraws to Near blockchain, and user never interacted with the token

If user never interacted with a token, then they need to pay NEP-141 storage. Which is deducted from withdraw amount, which is basically tiny (0.00125) swap to $NEAR. As it is a swap, then it's affected by slippage. This logic adds complexity: we can't receive less $NEAR than needed, because otherwise we couldn't afford storage and entire withdraw will fail. To solve it we need to account for slippage on amount in (amount of token to be withdrawn) rather than amount out (amount of $NEAR). But it adds another problem, we need know for sure how many tokens to will withdrawn. 

We decided not to account for slippage during withdrawals at all right now: 
1. natively multichain assets are 1-1
2. storage swaps are that insignificant, so solvers can increase deadline for such swaps

# Follow-up

1. Allow worse quotes that are within slippage, right now worse quotes are discarded even if user signed a message with less amount.  https://github.com/defuse-protocol/defuse-sdk/pull/277
2. Display slippage and min amount out https://github.com/defuse-protocol/defuse-sdk/pull/276
3. Add configration of slippage from settings

### DIFF
--- a/src/utils/tokenUtils.test.ts
+++ b/src/utils/tokenUtils.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from "vitest"
 import type { BaseTokenInfo, UnifiedTokenInfo } from "../types/base"
 import {
   DuplicateTokenError,
-  accountSlippage,
+  accountSlippageExactIn,
   addAmounts,
   adjustDecimals,
   compareAmounts,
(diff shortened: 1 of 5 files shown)


src/utils/tokenUtils.test.ts
```diff
@@ -883,3 +884,61 @@ describe("getUnderlyingBaseTokenInfos", () => {
...
+    ])
+  })
+
+  it("should handle slippage of 100%", () => {
+    const delta: Delta = [["token1", 1000n]]
+    const slippageBasisPoints = 10000 // 100%
+    expect(accountSlippageExactIn(delta, slippageBasisPoints)).toEqual([
+      ["token1", 0n],
+    ])
+  })
```
jobotics, line 943: Some cases that could also be covered:
1. should handle rounding correctly for small numbers
2. should handle very large numbers
3. should apply slippage to multiple positive amounts

jobotics (approved):

PR #277, opened 2025-02-03
feat: consider slippage when choosing the best quote

cawabunga-bytes:
How the best quote used to be determined:
1. We set initial quote
2. User starts signing a message with amount out from the initial quote
3. All coming quotes are compared to the initial quote: if a new one is better, then choose it
4. User finishes singing
5. Publish intent

It's straight forward approach, we store only two quotes.

Slippage introduces complexity and we can't do the same. Because we may end up with worse rates for the user.
Instead, we need to store all new quotes and at the moment of publishing the intent, we select the best quote.
So new flow will be:
1. (same) We set initial quote
2. (same) User starts signing a message with amount out from the initial quote
3. All coming quotes are compared to the initial quote adjusted to slippage: if a new one within slippage, then we store it in the queue, otherwise we discard it (as it is useless and we can't use it at all)
4. (same) User finishes singing
5. Choose the best quote from the queue that is valid for next 10 seconds
6. (same) Publish intent

### DIFF
--- a/src/features/machines/swapIntentMachine.ts
+++ b/src/features/machines/swapIntentMachine.ts
@@ -78,7 +78,6 @@ export type IntentDescription =
       type: "swap"
       totalAmountIn: TokenValue
       totalAmountOut: TokenValue
-      quote: AggregatedQuote
     }
   | {
       type: "withdraw"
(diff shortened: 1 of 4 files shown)


jobotics:
By point 5. in the new flow: What happens if this is optimistic intent within the balance in-flight, and the time before publishing takes several minutes?

cawabunga-bytes:
> By point 5. in the new flow: What happens if this is optimistic intent within the balance in-flight, and the time before publishing takes several minutes?

I don't know how optimistic balances will work, so I can't answer your question.

jobotics (approved):

PR #279, opened 2025-02-05
feat: split swap intent machine

jobotics:
This PR splits the intent handling business logic into two separate state machines to enhance accessibility for upcoming optimistic balance features.

The current architecture, where `swapIntentMachine` handles both signing and broadcasting, makes it difficult to support non-linear flows where some states may occur later.

## Key Changes:
1. Moved signing logic to `intentSignMachine` without behavioral changes
2. Moved broadcasting logic to `intentBroadcastMachine` without behavioral changes
3. Updated UI machines (`swapUIMachine`and `withdrawUIMachine`) to work with the new actors
4. Removed `swapIntentMachine`

Notes: The new machines introduce an extending state for swapping and withdrawal to handle broadcasting.

### Caveats: Extending actors causes type issues (TS7056 error) that currently have no solution.

Adds unit tests for swap and withdraw machine states.
Screenshot 2025-02-13 at 11 17 42


src/features/machines/intentBroadcastMachine.ts
```diff
@@ -0,0 +1,382 @@
...
+
+    if (context.error != null) {
+      return context.error
+    }
+
+    throw new Error("Unexpected output")
+  },
+
+  on: {
+    NEW_QUOTE: {
```
cawabunga-bytes: I thought broadcastMachine just publishes an intent. Quote should be known before spawning this machine, and no subscription to new quote is needed, because this machine sends the request right away. Am I missing something?

src/features/machines/intentBroadcastMachine.ts
```diff
@@ -0,0 +1,359 @@
...
+  userAddress: string
+  userChainType: ChainType
+  defuseUserId: DefuseUserId
+  referral?: string
+  slippageBasisPoints: number
+  nearClient: providers.Provider
+  sendNearTransaction: SendNearTransaction
+  intentOperationParams: IntentOperationParams
+  quoteToPublish: AggregatedQuote | null
+  quotes: PriorityQueue<AggregatedQuote>
```
cawabunga-bytes: You don't need this for publishing, you need just a quote.

jobotics:
Outdated code.

PR #281, opened 2025-02-07
feat: add pool

jobotics:
This PR introduces a new abstraction for handling execution intents within an extended flow in the intent pool. These changes enable us to manage execution intents, such as broadcasting, with specified execution logic. For instance, during execution when a deposit balance is still in-flight (optimistic balances), we can hold the intent in the intent pool and execute it non-sequentially.

This feature is backward compatible as the intent pool does not interfere with the existing intent workflow. It can be enabled or disabled using a feature flag.

## Key Changes:
1. Extend Settings: Added a new key `optimisticBalanceUpdates` (currently set to off).
2. Add intentPoolMachine: Introduced a new machine to manage intents within the pool.
3. Extend Swap Flow: Enhanced the swap flow within the pool state.
4. Extend Withdraw Flow: Enhanced the withdraw flow within the pool state.

This PR depends on #279.


jobotics:
Archived.

PR #285, opened 2025-02-18
feat: add rpc failover in NEAR runtime

jobotics:


src/services/nearHttpClient/runtime.ts
```diff
@@ -26,12 +26,15 @@ async function request(url: string, body: unknown): Promise<Response> {
 export async function jsonRPCRequest<
   T extends types.JSONRPCRequest<unknown, unknown>,
 >(method: T["method"], params: T["params"][0]) {
-  const response = await request(`${BASE_URL}`, {
-    id: "dontcare",
-    jsonrpc: "2.0",
-    method,
-    params: params !== undefined ? params : undefined,
-  })
+  const response = await failover(settings.reserveRpcUrls.near, (url) =>
```
jobotics: This is the only reason I decided to keep `failover`. Currently, we use `jsonRPCRequest` for several requests: getNearBalance, getNearNep141BalanceAccount, getTx, getNearNep141StorageBalanceOf, and getNearNep141StorageBalanceBounds.

This could probably be rewritten to nearClient somehow, but I'm still unsure whether we need to do so.

src/features/swap/components/SwapSubmitter.tsx
```diff
@@ -32,8 +33,8 @@ export function SwapSubmitterProvider({
       params: {
         userAddress,
         userChainType,
-        nearClient: new providers.JsonRpcProvider({
-          url: "https://rpc.mainnet.near.org",
+        nearClient: nearFailoverRpcProvider({
```
cawabunga-bytes: I think we need to create a global instance of the provider and just import it everywhere

cawabunga-bytes (approved):

dr-block (approved):

PR #324, opened 2025-03-05
fix: update OTC fee calculation

cawabunga-bytes:
In both OTC forms, we show approximate amounts, which doesn't represent actual breakdown. Because trade breakdown depends on user's balance. This PR addresses this issue and incorrect fee computation issue, also includes required refactoring. Pr can be reviewed by commits.

### DIFF
--- a/src/features/otcDesk/components/OtcMakerTrades.tsx
+++ b/src/features/otcDesk/components/OtcMakerTrades.tsx
@@ -45,9 +45,8 @@ import {
 } from "../stores/otcMakerTrades"
 import type { SignMessage } from "../types/sharedTypes"
 import {
-  type DetermineTokenInAndOutErr,
-  determineTokenInAndOut,
-  getTokenIds,
+  type DetermineInvolvedTokensErr,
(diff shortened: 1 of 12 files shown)


src/features/otcDesk/components/OtcMakerTrades.tsx
```diff
@@ -127,22 +126,16 @@ function OtcMakerTradeItem({
...
-    .andThen((tradeTerms) => {
-      const { tokenIdsIn, tokenIdsOut } = getTokenIds(tradeTerms.tokenDiff)
-      const tokensResult = determineTokenInAndOut(
-        tokenList,
-        tokenIdsIn,
-        tokenIdsOut
-      )
-      return tokensResult.map(({ tokenIn, tokenOut }) => {
-        return {
+    .mapErr<ParseTradeTermsErr | DetermineInvolvedTokensErr>((a) => a)
```
jobotics, line 129: What is a purpose of this `.mapErr`  as it's just mapping the error to itself without any transformation?

PR #326, opened 2025-03-07
feat: gift

jobotics:
## Feature: Gift
This feature enables secure token transfers to a temporary escrow account with the ability to reclaim tokens without requiring wallet events from the requester side.

## How it works
Prerequisites: We utilize ED25519 for enhanced flexibility within the NEAR ecosystem
1. Generate an ED25519 keypair for the escrow account (tempUser)
2. Transfer tokenA to the tempUser account
3. Create a secret link containing the private key for tempUser
4. Enable tokenA to be claimed by claimerUser using the secret from the link

### Technical Notes
- This implementation is safe to merge as it operates independently of existing logic


src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
+import { BorshSchema, borshSerialize } from "borsher"
+
+export enum SignStandardEnum {
+  nep413 = "nep413",
+}
+
+const standardNumber = {
+  [SignStandardEnum.nep413]: 413,
```
cawabunga-bytes: I don't think we'll support other standards (if there is any). So it can be removed

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+ * The resulting hash should be used with EdDSA signing to create a valid NEP-413 signature.
+ * Note: This is a browser-only implementation using Web Crypto API.
+ *
+ * @param intentMessage - Message content to be signed
+ * @param recipient - Recipient account ID
+ * @param nonce - Base64 encoded 32-byte nonce
+ * @param standard - Signature standard (currently only NEP-413)
+ * @returns Promise resolving to Buffer containing message hash for signing
+ */
+export async function serializeIntent(
```
cawabunga-bytes: File is called `hashing`, but function `serialize`? 
You hash any message, so I wouldn't stick to "intent".

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,61 @@
...
+
+type GiftToken = {
+  tokenInDiff: Record<BaseTokenInfo["defuseAssetId"], bigint>
+  tokenIn: BaseTokenInfo
+}
+
+export async function determineGiftToken(
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[],
+  giftSecret: GiftSecret
+): Promise<Result<GiftToken, string>> {
```
cawabunga-bytes: Please explicitly specify errors

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { GiftSecret } from "./parseGiftSecret"
+
+type GiftToken = {
+  tokenInDiff: Record<BaseTokenInfo["defuseAssetId"], bigint>
+  tokenIn: BaseTokenInfo
+}
+
+export async function determineGiftToken(
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[],
+  giftSecret: GiftSecret
```
cawabunga-bytes: It's easier to just pass user id.

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const tokenIn = getAnyBaseTokenInfo(tokenIn_)
+
+    return Ok({
+      tokenInDiff,
+      tokenIn,
+    })
+  } catch (error) {
+    return Err(
+      `ERR_GETTING_BALANCES: ${error instanceof Error ? error.message : String(error)}`
```
cawabunga-bytes: Try-catch only part which may throw. Also such error string is kind of bad, how would you type it?

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+const standardNumber = {
+  [SignStandardEnum.nep413]: 413,
+}
+
+interface ITokenDiff {
+  intent: "token_diff"
+  diff: { [key: string]: string }
+}
+
+type IIntent = ITokenDiff
```
cawabunga-bytes: That's redundant.

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Message structure for signing
+ */
+export interface IMessage {
```
cawabunga-bytes: No need export

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+
+interface ITokenDiff {
+  intent: "token_diff"
+  diff: { [key: string]: string }
+}
+
+type IIntent = ITokenDiff
+
+/**
+ * Message structure for signing
```
cawabunga-bytes: Not useful comment

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export interface IMessage {
+  signer_id: string
+  deadline: string
+  intents: IIntent[]
+}
+
+/**
+ * Borsh schema for NEP-413 payload serialization
```
cawabunga-bytes: Not useful comment too, you can just read `const nep413PayloadSchema = BorshSchema.Struct`

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+  callback_url: BorshSchema.Option(BorshSchema.String),
+})
+
+/**
+ * Client-side utility to serialize and hash NEP-413 messages for EdDSA signature verification.
+ * Follows the NEP-413 specification for message serialization and hashing:
+ * @see https://github.com/near/NEPs/blob/master/neps/nep-0413.md#specification
+ *
+ * The resulting hash should be used with EdDSA signing to create a valid NEP-413 signature.
+ * Note: This is a browser-only implementation using Web Crypto API.
```
cawabunga-bytes: This is not true. Nodejs also supports web crypto api. But browsers don't have `Buffer`, so your comment is opposite. It should not return Buffer.

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+  intentMessage: unknown,
+  recipient: string,
+  nonce: string,
+  standard: SignStandardEnum
+): Promise<Buffer> {
+  if (!standardNumber[standard]) {
+    throw new Error(`Unsupported standard: ${standard}`)
+  }
+
+  // Prepare payload with fixed-size nonce array
```
cawabunga-bytes: This is not useful

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+  const payload = {
+    message: intentMessage,
+    nonce: new Uint8Array(32),
+    recipient,
+  }
+  const nonceData = Buffer.from(nonce, "base64")
```
cawabunga-bytes: Please Use `base64.decode` from `@scure/base` package

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,34 @@
...
+    return Err("CANNOT_PARSE_SECRET_KEY")
+  }
+
+  return Ok({
+    secretKey: parseResult.output,
+    walletId: deriveWalletId(parseResult.output),
+  })
+}
+
+const SecretKeyPlainSchema = v.string()
```
cawabunga-bytes: You don't need this, you already have `string` type.

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,79 @@
...
+  payload.nonce.set(nonceData.subarray(0, 32))
+
+  // Serialize payload and combine with standard identifier
+  const payloadSerialized = borshSerialize(nep413PayloadSchema, payload)
+  const baseInt = 2 ** 31 + standardNumber[standard]
+  const baseIntSerialized = borshSerialize(BorshSchema.u32, baseInt)
+  const combinedData = Buffer.concat([baseIntSerialized, payloadSerialized])
```
cawabunga-bytes: Please don't use Buffer, because it's not browser standard, I don't even know where it comes from.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function deriveWalletId(secretKeyBase58: string): string {
+  const secretKey = bs58.decode(secretKeyBase58)
+  const keyPair = sign.keyPair.fromSecretKey(secretKey)
+  return hex.encode(keyPair.publicKey)
```
cawabunga-bytes: BTW this is not always correct for Near blockchain. Here is a scenario:
1. Generate a keypair (1)
2. Generate another keypair (2)
3. Add public key (2) to account (1)
4. Remove public key (1) from the account
5. Now you have account id from keypair (1), but private key from keypair (2)

Personally I did it multiple times

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  const parseResult = v.safeParse(SecretKeyPlainSchema, secretKey)
+  if (!parseResult.success) {
+    return Err("CANNOT_PARSE_SECRET_KEY")
+  }
+
+  return Ok({
+    secretKey: parseResult.output,
+    walletId: deriveWalletId(parseResult.output),
```
cawabunga-bytes: This function will throw if you passed incorrect base58 string.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,34 @@
...
+import { sign } from "tweetnacl"
+import * as v from "valibot"
+
+export type GiftSecret = {
+  secretKey: string
+  walletId: string
+}
+
+export function parseGiftSecret(secretKey: string): Result<GiftSecret, string> {
+  if (secretKey.length === 0) {
```
cawabunga-bytes: This doesn't look useful

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+import { SignStandardEnum, serializeIntent } from "./hashing"
+
+export async function signGiftTakerMessage({
+  giftTerms,
+  signerCredentials,
+}: {
+  giftTerms: GiftTerms
+  signerCredentials: SignerCredentials | null
+}): Promise<Result<NEP413SignatureData, string>> {
+  if (signerCredentials == null) {
```
cawabunga-bytes: That's weird, you can skip this check just by specifying valid argument type `signerCredentials: SignerCredentials` (without null)

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+    return Err("CREDENTIALS_NOT_FOUND")
+  }
+  const nonce = randomDefuseNonce()
+
+  const innerMessage = makeInnerTransferMessage({
+    tokenDeltas: [...Object.entries(giftTerms.tokenDiff)],
+    signerId: resolveSignerId(giftTerms.walletId as DefuseUserId),
+    deadlineTimestamp: minutesFromNow(5),
+    receiverId: signerCredentials.credential,
+    memo: "GIFT_FILL",
```
cawabunga-bytes: `GIFT_FILL` it doesn't make sense for me :) Maybe claim?

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+  })
+  const walletMessage = makeSwapMessage({
+    innerMessage,
+    nonce: nonce,
+  })
+
+  try {
+    const keyPair = KeyPair.fromString(`ed25519:${giftTerms.secretKey}`)
```
cawabunga-bytes: Curve may be included to string itself, so you don't need to hardcode it. Also it will be future proof, if we decide to use different type of keypair, for example, ECDSA (Ethereum).

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const serialize = await serializeIntent(
+      walletMessage.NEP413.message,
+      walletMessage.NEP413.recipient,
+      base64.encode(nonce),
```
cawabunga-bytes: Unnecessary work, you encode here and then decode in the function. You can just pass raw bytes.

Also I noticed, you can just pass entire `walletMessage.NEP413` as it contains everything you need, it's called `NEP413` for the reason.

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+        signature: base64.encode(signature.signature),
+      },
+      signedData: walletMessage.NEP413,
+    })
+  } catch (error) {
+    return Err(
+      `Failed to sign message: ${error instanceof Error ? error.message : String(error)}`
+    )
+  }
+}
```
cawabunga-bytes, line 67: This function does two things: 
1. assemble wallet message to sign
2. nep413 sign

So you can split your function into two.

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,75 @@
...
+      signatureData: {
+        accountId: giftTerms.walletId,
+        publicKey: keyPair.getPublicKey().toString(),
+        signature: base64.encode(signature.signature),
+      },
+      signedData: walletMessage.NEP413,
+    })
+  } catch (error) {
+    return Err(
+      `Failed to sign message: ${error instanceof Error ? error.message : String(error)}`
```
cawabunga-bytes: This is unparsable error, please avoid generic errors.

src/utils/messageFactory.ts
```diff
@@ -313,3 +313,36 @@ export function makeChallenge(payload: Uint8Array): Uint8Array {
...
+  tokenDeltas: [string, bigint][]
+  signerId: DefuseUserId
+  deadlineTimestamp: number
+  receiverId: string
+  memo?: string
+}): Nep413DefuseMessageFor_DefuseIntents {
+  const tokens: Record<string, string> = {}
+
+  for (const [token, amount] of tokenDeltas) {
+    tokens[token] = amount.toString()
```
cawabunga-bytes, line 340: `tokenDeltas` is array, so it can contain duplicated tokens. Such overriding may cause unexpected behaviour and fund loss. Also since it's called `delta` I assume it can be negative, your code doesn't handle it.

src/features/gift/components/GiftTakerForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+  signerCredentials: SignerCredentials | null
+  onSuccessClaim: (arg: { intentHashes: string[] }) => void
+}
+
+export function GiftTakerForm({
+  giftTerms,
+  signerCredentials,
+  onSuccessClaim,
+}: GiftTakerFormProps) {
+  const amountIn = computeTotalBalanceDifferentDecimals(
```
cawabunga-bytes: Not `amountIn`, just amount, please fix in other places too.

src/features/gift/hooks/useGiftTakerConfirmClaim.ts
```diff
@@ -0,0 +1,26 @@
+import { useMutation } from "@tanstack/react-query"
+import {
+  type SignerCredentials,
+  formatSignedIntent,
+} from "../../../core/formatters"
+import { publishIntents } from "../../../services/intentService"
+import type { NEP413SignatureData } from "../../../types/swap"
+
+export function useGiftTakerConfirmClaim() {
```
cawabunga-bytes: We don't need to use mutations at all, since it should be handled in machines. I used mutations because I didn't use machines, because of time constraints.

src/features/gift/utils/deriveGiftTerms.ts
```diff
@@ -0,0 +1,36 @@
...
+import { parseGiftSecret } from "./parseGiftSecret"
+
+export type GiftTerms = {
+  tokenDiff: Record<BaseTokenInfo["defuseAssetId"], bigint>
+  tokenIn: BaseTokenInfo | UnifiedTokenInfo
+  secretKey: string
+  walletId: string
+}
+
+export async function deriveGiftTerms(
```
cawabunga-bytes: It's `getGiftInfo`, because our gifts don't have terms.

src/features/gift/components/ShareableGiftImage.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export function ShareableGiftImage({
+  token,
+  amountIn,
+  message,
+  className,
+}: ShareableGiftImageProps) {
+  return (
+    <div
+      className={`relative w-full min-w-[334.22px] min-h-[188px] max-w-[600px] h-auto aspect-[1.9/1] rounded-xl flex flex-col justify-center p-10 items-center ${className}`}
```
cawabunga-bytes: We have `src/utils/cn.ts` that handles tailwind classes overriding

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [giftTerms, setGiftTerms] = useState<Result<GiftTerms, string> | null>(
+    null
+  )
+
+  const [claimResult, setClaimResult] = useState<{
+    intentHashes: string[]
+  } | null>(null)
+
+  useEffect(() => {
+    deriveGiftTerms(secretKey, tokenList).then((result) => {
```
cawabunga-bytes: It should be in the machine

src/features/gift/components/GiftMakerReadyDialog.tsx
```diff
@@ -0,0 +1,200 @@
...
+  readyGiftRef: ActorRefFrom<typeof giftMakerReadyActor>
+  signerCredentials: SignerCredentials
+  signMessage: SignMessage
+  generateLink: (secretKey: string) => string
+}
+
+export function GiftMakerReadyDialog({
+  readyGiftRef,
+  // biome-ignore lint/correctness/noUnusedVariables: <explanation>
+  signerCredentials,
```
cawabunga-bytes: Please remove unused variables.

cawabunga-bytes (changes requested):

src/core/messages.test.ts
```diff
@@ -93,3 +94,27 @@ describe("createEmptyIntentMessage()", () => {
...
+      memo: "message",
+      deadlineTimestamp: TEST_TIMESTAMP,
+    })
+
+    expect(JSON.parse(message.NEP413.message)).toEqual({
+      deadline: new Date(TEST_TIMESTAMP).toISOString(),
+      intents: [
+        {
+          intent: "transfer",
+          tokens: { "token.near": "-100" },
```
cawabunga-bytes: wut, negative transfer?

src/features/gift/actors/giftMakerEscrowActor.ts
```diff
@@ -0,0 +1,46 @@
...
+import { sign } from "tweetnacl"
+import { setup } from "xstate"
+
+export type EscrowKeyPair = {
+  publicKey: string
+  secretKey: string
+  walletId: string
+}
+
+export const giftMakerEscrowActor = setup({
```
cawabunga-bytes: Machine does nothing.

src/features/gift/actors/giftMakerCancellationActor.ts
```diff
@@ -0,0 +1,361 @@
...
+type GiftMakerCancellationActorContext = {
+  giftId: string
+  secretKey: string
+  signerCredentials: SignerCredentials
+  multiPayload: MultiPayload
+  tokenIn: BaseTokenInfo | UnifiedTokenInfo
+  error: null | GiftMakerCancellationActorErrors
+}
+
+export const giftMakerCancellationActor = setup({
```
cawabunga-bytes: I don't get this, it should be the same as claiming the gift. Is it just copy-paste or it has different logic?

cawabunga-bytes:
Please provide follow-up plan, like what's missing and what needs to be fixed.

src/features/gift/utils/hashing.ts
jobotics: That's a good comment, thanks for noticing.

src/features/gift/utils/determineGiftToken.ts
jobotics: Agree, I'll remove a redundant part.

src/features/gift/utils/parseGiftSecret.ts
jobotics: Possibly yes, but it's probably never the case unless we do it ourselves or the user does it manually by mistake. In either case, I'm not sure how we can handle it.

cawabunga-bytes: In our case is probably yes, as we control the key and lifespan is low. We don't want to re-use this function elsewhere.

src/features/gift/actors/giftMakerCancellationActor.ts
jobotics: At the moment they are different, when we claim from the taker side, we extract data from the secret, in the maker side, we take it from the payload that we were signed, so `signing` state is a bit different.

For comparison: 
- maker signing - https://github.com/defuse-protocol/defuse-sdk/blob/feat/gift/src/features/gift/actors/giftMakerCancellationActor.ts#L85
- taker signing - https://github.com/defuse-protocol/defuse-sdk/blob/feat/gift/src/features/gift/actors/giftTakerRootMachine.ts#L115

We can put it in one `claim` actor but we are constraint in time.

jobotics:
> Please provide follow-up plan, like what's missing and what needs to be fixed.

Missing / To-Do Items:
1. Implement persistent storage for giftMakerStore across multiple levels:
- LocalStorage
- SessionStorage
- IndexedDB
2. Add Pending section
3. Add History section
4. Fix social media preview images

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+    token: BaseTokenInfo | UnifiedTokenInfo | undefined
+  }>(ModalType.MODAL_SELECT_ASSETS, "token")
+
+  const updateTokens = useTokensStore((state) => state.updateTokens)
+
+  const handleSelect = (fieldName: string) => {
+    updateTokens(tokenList)
+    setModalType(ModalType.MODAL_SELECT_ASSETS, {
+      fieldName,
+      selectToken: undefined,
```
dr-block, line 115: NIT: after your merge I should come and update the logic here, as I am going to change a bit modal store logic and below useEffect

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+        readyGiftRef != null &&
+        signerCredentials != null && (
+          <GiftMakerReadyDialog
+            readyGiftRef={readyGiftRef}
+            generateLink={generateLink}
+          />
+        )}
+
+      {/* Header Section */}
+      <div className="flex flex-row justify-between mb-5">
```
dr-block: can this go to separate component ? let's make this component as much as possible smaller

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+          </div>
+        </div>
+
+        {renderSubmitButton(rootSnapshot)}
+      </form>
+    </div>
+  )
+}
+
+function renderSubmitButton(
```
dr-block: why keep it here in the same component ? why not haveing SubmitButton component with prop rootSnapshot ?

src/features/gift/components/GiftMakerReadyDialog.tsx
```diff
@@ -0,0 +1,194 @@
...
+import type { giftMakerCancellationActor } from "../actors/giftMakerCancellationActor"
+import type { giftMakerReadyActor } from "../actors/giftMakerReadyActor"
+import { ShareableGiftImage } from "./ShareableGiftImage"
+
+type GiftMakerReadyDialogProps = {
+  readyGiftRef: ActorRefFrom<typeof giftMakerReadyActor>
+  generateLink: (secretKey: string) => string
+}
+
+export function GiftMakerReadyDialog({
```
dr-block, line 25: what if we have in separate files GiftMakerReadyDialog and CancellationDialog?

src/features/gift/components/GiftMakerReadyDialog.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const finish = () => {
+    readyGiftRef.send({ type: "FINISH" })
+  }
+
+  const cancelGift = () => {
+    readyGiftRef.send({ type: "CANCEL_ORDER" })
+  }
+
+  return (
+    <BaseModalDialog open={true} onClose={finish} isDismissable>
```
dr-block: guess open={true} can be changed to just open

src/features/gift/components/GiftMakerReadyDialog.tsx
```diff
@@ -0,0 +1,194 @@
...
+          )}
+
+          <div className="flex flex-col md:flex-row justify-center gap-3 mt-5">
+            <Button
+              type="button"
+              size="4"
+              variant="outline"
+              color="gray"
+              className="md:flex-1 font-bold"
+              onClick={() => actorRef?.send({ type: "ABORT_CANCELLATION" })}
```
dr-block: we can have parametral function like sendType with useCallbakc wrapper 
sendType(type)
and call that for ABORT_CANCELLATION, CONFIRM_CANCELLATION, ACK_CANCELLATION_IMPOSSIBLE ...

src/features/gift/components/GiftTakerForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+  giftTakerClaimRef,
+}: GiftTakerFormProps) {
+  const amount = computeTotalBalanceDifferentDecimals(
+    getUnderlyingBaseTokenInfos(giftInfo.token),
+    giftInfo.tokenDiff,
+    { strict: false }
+  )
+
+  const snapshot = useSelector(giftTakerClaimRef, (state) => state)
+  const processing = snapshot?.matches("claiming")
```
dr-block, line 62: should not "claiming" be somewhere declared already, can we have reussble constants for this ?

src/features/gift/components/GiftTakerForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+  assert(amount != null)
+
+  return (
+    <div className="flex flex-col">
```
dr-block, line 66: in general this can be totally separate component

src/features/gift/components/GiftTakerForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+      </div>
+
+      {/* Image Section */}
+      <ShareableGiftImage
+        token={giftInfo.token}
+        amount={amount}
+        message="You've received a gift! Click to claim it."
+      />
+
+      {snapshot?.context.error != null && (
```
dr-block: this looke separate component as well with it's logic

src/features/gift/components/GiftTakerSuccessScreen.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const txUrl =
+    intentStatus.data?.txHash != null
+      ? `${NEAR_EXPLORER}/txns/${intentStatus.data.txHash}`
+      : null
+
+  return (
+    <div>
+      {/* Header Section */}
+      <div className="flex flex-row justify-between mb-5">
```
dr-block: still separate component ?

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -0,0 +1,87 @@
+import { useActorRef, useSelector } from "@xstate/react"
+import {} from "react"
```
dr-block: think can be removed, below code as well, getGiftInfo as well

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -0,0 +1,87 @@
...
+      : null
+
+  const snapshot = useSelector(giftTakerClaimRef, (state) => state)
+
+  if (snapshot?.context.error != null) {
+    return <GiftTakerInvalidClaim error={snapshot.context.error.reason} />
+  }
+
+  if (snapshot.context.giftInfo == null) {
+    return loading
```
dr-block, line 65: just string loading ? may be any spam wrapper ?

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  /** List of available tokens for trading */
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[]
+
+  /** User's wallet address */
+  userAddress: string | null | undefined
+  userChainType: ChainType | null | undefined
+
+  /** Theme selection */
+  theme?: "dark" | "light"
```
dr-block, line 23: think I saw this already, how about having Theme type?

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -0,0 +1,87 @@
...
+    return <GiftTakerInvalidClaim error={snapshot.context.error.reason} />
+  }
+
+  if (snapshot.context.giftInfo == null) {
+    return loading
+  }
+
+  return (
+    <>
+      {snapshot.status === "done" && snapshot.context.intentHashes ? (
```
dr-block: still I would prefer to know from where this string "done" comes, any constant reference ?

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+      return isBaseToken(token)
+        ? [token.defuseAssetId]
+        : token.groupedTokens.map((t) => t.defuseAssetId)
+    })
+
+    const balances = await getDepositedBalances(
+      userAddressToDefuseUserId(userId, "near"),
+      tokenIds,
+      new providers.JsonRpcProvider({
+        url: "https://nearrpc.aurora.dev",
```
dr-block, line 39: may be this is related to env, I guess we should take from somewhere this, not hardcode

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+): Promise<Result<GiftToken, DetermineGiftTokenErr>> {
+  try {
+    const tokenIds = tokenList.flatMap((token) => {
+      return isBaseToken(token)
+        ? [token.defuseAssetId]
+        : token.groupedTokens.map((t) => t.defuseAssetId)
+    })
+
+    const balances = await getDepositedBalances(
+      userAddressToDefuseUserId(userId, "near"),
```
dr-block: any const isntead "near"?

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,57 @@
+import { hex } from "@scure/base"
+import { Err, Ok, type Result } from "@thames/monads"
+import bs58 from "bs58"
+import { sign } from "tweetnacl"
+import * as v from "valibot"
```
dr-block, line 2: can we import just those which we need, not all

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,57 @@
...
+  const keyPair = sign.keyPair.fromSecretKey(secretKeyBase58)
+  return hex.encode(keyPair.publicKey)
+}
+
+export function deriveSecretKey(
+  escrowCredentials: EscrowCredentials,
+  standard: "nep413"
+): string {
+  switch (standard) {
+    case "nep413":
```
dr-block: i saw nep413 in a lot of places ... any const that we can make and reuse ?

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,57 @@
...
+  try {
+    const parseResult = v.safeParse(v.string(), secretKey)
+    if (!parseResult.success) {
+      return Err("CANNOT_PARSE_SECRET_KEY")
+    }
+
+    return Ok({
+      secretKey: parseResult.output,
+      // If we support different escrow accounts in the future, we should distinguish how to normalize them
+      userId: deriveUserId(normalizeNEP413Key(parseResult.output)),
```
cawabunga-bytes: It's not userId, it is account id of the escrow. If you want to return user id then you need to use `userAddressToDefuseUserId`. It's proper place for doing it, instead of parsing it somewhere else (e.g. src/features/gift/utils/determineGiftToken.ts).

src/features/gift/actors/giftMakerCancellationActor.ts
```diff
@@ -0,0 +1,364 @@
...
+      }) => {
+        // We need to extract tokenDiff from the original gift intent to reclaim it
+        if (input.multiPayload.standard !== "nep413") {
+          return {
+            tag: "err" as const,
+            value: { reason: "CANNOT_PARSE_GIFT_MULTIPAYLOAD" as const },
+          }
+        }
+
+        const giftMultiPayload = JSON.parse(input.multiPayload.payload.message)
```
cawabunga-bytes: You can use `MultiPayloadDeepSchema` to parse entire multipayload object.

src/features/gift/utils/parseGiftSecret.ts
cawabunga-bytes: I checked it, valibot is tree-shakable, so such import is fine. I wouldn't mind to import only needed parts, but it becomes very inconvenient to use.

src/features/gift/utils/generateEscrowCredentials.ts
```diff
@@ -0,0 +1,33 @@
+import { base58, hex } from "@scure/base"
+import { sign } from "tweetnacl"
+
+export type EscrowCredentials = {
+  NEP413: {
```
cawabunga-bytes: I think this is misused. NEP-413 is just a standard for singing a message. For example, Ethereum has ERC-191 and EIP-712 standards. In this case it is just EdDSA keypair.

src/features/gift/utils/generateEscrowCredentials.ts
```diff
@@ -0,0 +1,33 @@
...
+    standard: "nep413"
+    secretKey: string
+    userId: string
+  }
+}
+
+export function generateEscrowCredentials(): EscrowCredentials {
+  const keyPair = sign.keyPair()
```
cawabunga-bytes, line 10: Curious why did you choose to generate a key using tweetnacl instead of near-api-js?

src/features/gift/components/GiftTakerWidget.tsx
jobotics: Oh, yes, linter remove unused variable and left)

src/features/gift/components/GiftMakerForm.tsx
jobotics: Not really, for example at `src/features/gift` repository we have 4 such sections and each of them has some difference.

src/features/gift/components/GiftMakerForm.tsx
jobotics: We use ButtonCustom so in this case we just wrapup logic for button caption and loading process

jobotics: Actually, it's redundant, we simply can use just ButtonCustom above

src/features/gift/components/GiftMakerReadyDialog.tsx
jobotics: Not sure as they are both steak to `readyGiftRef` actor, meaning they can't work separately, so in case of separation they also have to be grouped somehow.

src/features/gift/utils/determineGiftToken.ts
jobotics: This has to be changed with this pr - https://github.com/defuse-protocol/defuse-sdk/pull/285

src/features/gift/utils/determineGiftToken.ts
jobotics: We use it only once so `const` will give nothing. NIT: we use `near` escrow account for Gift so we can state that it's true

src/features/gift/utils/parseGiftSecret.ts
jobotics: Good point, but I think it needs to be done separately, since we use the standard as a string not only here

src/features/gift/components/GiftTakerWidget.tsx
jobotics: It comes from `giftTakerRootMachine` machine when it is in `final` state

src/features/gift/components/GiftMakerForm.tsx
jobotics: Ok

src/features/gift/components/GiftTakerForm.tsx
jobotics: We don't have it before

src/features/gift/components/GiftTakerForm.tsx
jobotics: Answered above

src/features/gift/components/GiftTakerSuccessScreen.tsx
jobotics: Answered above

src/features/gift/components/GiftTakerWidget.tsx
jobotics: We don't have a design for this element, for now, it looks fine, but we can make it fancy later

src/features/gift/components/GiftTakerWidget.tsx
jobotics: Yeah, theme in our todo list

src/features/gift/actors/giftMakerCancellationActor.ts
cawabunga-bytes: Yes we always have time constraints, but we don't have strict deadline for this feature. So let's do it correctly.

src/features/gift/utils/generateEscrowCredentials.ts
```diff
@@ -0,0 +1,33 @@
+import { base58, hex } from "@scure/base"
+import { sign } from "tweetnacl"
+
+export type EscrowCredentials = {
+  NEP413: {
+    standard: "nep413"
+    secretKey: string
+    userId: string
```
cawabunga-bytes: Strictly speaking it's not userId, it will be a Near accountId. If you want it to become a userId then you need to use `userAddressToDefuseUserId()`.

I would consider writing this type like this:
```ts
interface EscrowCredentials extends SignerCredentials {
  secretKey: string
}
```

src/features/gift/utils/generateEscrowCredentials.ts
```diff
@@ -0,0 +1,33 @@
...
+    NEP413: {
+      standard: "nep413",
+      secretKey: transformNEP413Key(keyPair.secretKey),
+      userId: hex.encode(keyPair.publicKey),
+    },
+  }
+}
+
+function transformNEP413Key(key: Uint8Array): string {
+  return `ed25519:${base58.encode(key)}`
```
cawabunga-bytes, line 30: I like serialization with an algorithm prefix, because it will allow us to change algorithm without breaking format backward capability. For example, if we were to change it to Ethereum private key, then it would look like `secp256k1:key`.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+    })
+  } catch {
+    return Err("ACCOUNT_NOT_FOUND")
+  }
+}
+
+// This function should be updated if we start supporting different escrow accounts
+export function deriveAccountId(secretKey: string): string {
+  const normalizedSecretKey = normalizeNEP413Key(secretKey)
+  const secretKeyBase58 = bs58.decode(normalizedSecretKey)
```
cawabunga-bytes: Please use `import { base58 } from "@scure/base"` for decoding/encoding base58

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+    return Ok({
+      secretKey: parseResult.output,
+    })
+  } catch {
+    return Err("ACCOUNT_NOT_FOUND")
+  }
+}
+
+// This function should be updated if we start supporting different escrow accounts
+export function deriveAccountId(secretKey: string): string {
```
cawabunga-bytes: This code almost duplicate of src/features/gift/utils/generateEscrowCredentials.ts but instead of generating a key, it accepts it as an argument. I suggest composing them together.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+  const normalizedSecretKey = normalizeNEP413Key(secretKey)
+  const secretKeyBase58 = bs58.decode(normalizedSecretKey)
+  const keyPair = sign.keyPair.fromSecretKey(secretKeyBase58)
+  return hex.encode(keyPair.publicKey)
+}
+
+export function deriveSecretKey(
+  escrowCredentials: EscrowCredentials,
+  standard: "nep413"
```
cawabunga-bytes: Additionally to my comment above about misuse of nep413, EscrowCredentials should already contain info about the algorithm (it is a prefix of the secret key), so you don't need to duplicate it.

cawabunga-bytes (changes requested):

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param standard - Signature standard (currently only NEP-413)
+ * @returns Promise resolving to Buffer containing message hash for signing
+ */
+export async function hashing(
+  intentMessage: unknown,
+  recipient: string,
+  nonce: Uint8Array,
+  standard: number
+): Promise<Uint8Array> {
+  if (standard !== 413) {
```
cawabunga-bytes: I pretty sure that we won't implement other singing standards. I suggest getting rid of `standard` argument and rename the function to `signMessageNEP413`

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Note: This is a browser-only implementation.
+ *
+ * @param intentMessage - Message content to be signed
+ * @param recipient - Recipient account ID
+ * @param nonce - Base64 encoded 32-byte nonce
+ * @param standard - Signature standard (currently only NEP-413)
+ * @returns Promise resolving to Buffer containing message hash for signing
+ */
+export async function hashing(
+  intentMessage: unknown,
```
cawabunga-bytes: It should be string or bytes, but definitely not unknown. Borsch schema says it is `message: BorshSchema.String`. Also it is just `message`.

src/features/gift/utils/hashing.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (standard !== 413) {
+    throw new Error(`Unsupported standard: ${standard}`)
+  }
+
+  const payload = {
+    message: intentMessage,
+    nonce: new Uint8Array(32),
+    recipient,
+  }
+  payload.nonce.set(nonce.subarray(0, 32))
```
cawabunga-bytes: You can do just `nonce: nonce,`, because you're not mutating the array.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
+import { hex } from "@scure/base"
+import { Err, Ok, type Result } from "@thames/monads"
+import bs58 from "bs58"
```
cawabunga-bytes: Here the same, please use "@scure/base"

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+import { sign } from "tweetnacl"
+import { safeParse, string } from "valibot"
```
cawabunga-bytes: Let's stick to `import * as v from "valibot"`, as it will be correctly tree-shakable

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+export type GiftSecretError =
+  | "SECRET_KEY_EMPTY"
+  | "CANNOT_PARSE_SECRET_KEY"
+  | "ACCOUNT_NOT_FOUND"
+
+export function parseGiftSecret(
+  secretKey: string
+): Result<GiftSecret, GiftSecretError> {
+  try {
+    const parseResult = safeParse(string(), secretKey)
```
cawabunga-bytes: This parsing is useless, because secretKey is already a string.

src/features/gift/utils/parseGiftSecret.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!parseResult.success) {
+      return Err("CANNOT_PARSE_SECRET_KEY")
+    }
+
+    return Ok({
+      secretKey: parseResult.output,
+    })
+  } catch {
```
cawabunga-bytes, line 38: try-catch is useless, because nothing is throwing in the block

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,82 @@
...
+}: GiftTakerMessage): Promise<Result<NEP413SignatureData, string>> {
+  const walletMessage = assembleWalletMessage({ giftInfo, signerCredentials })
+
+  try {
+    // With different types of escrow accounts this should be updated
+    const keyPair = KeyPair.fromString(giftInfo.secretKey)
+    const messageHash = await hashing(
+      walletMessage.NEP413.message,
+      walletMessage.NEP413.recipient,
+      walletMessage.NEP413.nonce,
```
cawabunga-bytes: You can pass entire `walletMessage.NEP413` object, it already contains all needed info

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,82 @@
...
+      type: "NEP413",
+      signatureData: {
+        accountId: giftInfo.userId,
+        publicKey: keyPair.getPublicKey().toString(),
+        signature: base64.encode(signature.signature),
+      },
+      signedData: walletMessage.NEP413,
+    })
+  } catch {
+    return Err("CANNOT_SIGN_GIFT_TAKER_MESSAGE")
```
cawabunga-bytes: I don't find this message helpful. Which of the function call above is expected to throw?

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,82 @@
...
+
+function assembleWalletMessage({
+  giftInfo,
+  signerCredentials,
+}: GiftTakerMessage) {
+  const nonce = randomDefuseNonce()
+
+  const innerMessage = makeInnerTransferMessage({
+    tokenDeltas: [...Object.entries(giftInfo.tokenDiff)],
+    signerId: resolveSignerId(
```
cawabunga-bytes: `resolveSignerId` call is not needed, because `userAddressToDefuseUserId` already returns correct string

src/features/gift/utils/signGiftTakerMessage.ts
```diff
@@ -0,0 +1,82 @@
...
+      userAddressToDefuseUserId(giftInfo.userId, "near")
```
cawabunga-bytes: I think I already left a comment somewhere above about userId. It should've been transformed (typecasted) to DefuseUserId already, so you don't guess if it is "near" address or not. Because it cause a bug, if secret key algorithm is changed.

src/features/gift/utils/getGiftInfo.ts
```diff
@@ -0,0 +1,34 @@
...
+  tokenDiff: Record<BaseTokenInfo["defuseAssetId"], bigint>
+  token: BaseTokenInfo | UnifiedTokenInfo
+  secretKey: string
+  userId: string
+}
+
+export async function getGiftInfo(
+  secretKey: string,
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[]
+): Promise<Result<GiftInfo, string>> {
```
cawabunga-bytes: Could you type errors too?

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+): Promise<Result<GiftToken, DetermineGiftTokenErr>> {
+  try {
+    const tokenIds = tokenList.flatMap((token) => {
+      return isBaseToken(token)
+        ? [token.defuseAssetId]
+        : token.groupedTokens.map((t) => t.defuseAssetId)
+    })
+
+    const balances = await getDepositedBalances(
+      userAddressToDefuseUserId(accountId, "near"),
```
cawabunga-bytes: This already should've been DefuseUserId, because secret key algorithm can change

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function determineGiftToken(
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[],
+  accountId: string
+): Promise<Result<GiftToken, DetermineGiftTokenErr>> {
+  try {
+    const tokenIds = tokenList.flatMap((token) => {
+      return isBaseToken(token)
+        ? [token.defuseAssetId]
+        : token.groupedTokens.map((t) => t.defuseAssetId)
+    })
```
cawabunga-bytes: You can use `getUnderlyingBaseTokenInfos` from tokenUtils.ts and then get ids

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+        : token.groupedTokens.some(
+            (gt) => tokenDiff[gt.defuseAssetId] !== undefined
+          )
+    )
+
+    if (!token_) {
+      return Err("NO_TOKEN_OR_GIFT_HAS_BEEN_CLAIMED")
+    }
+
+    const token = getAnyBaseTokenInfo(token_)
```
cawabunga-bytes: This is not correct. Let's say you're gifted $TURBO. Gift maker sent $TURBO@Solana, but `getAnyBaseTokenInfo` will most likely return you $TURBO@Ethereum. They have different decimals, so you end up with incorrect data. So you can't just pick random token.

src/features/gift/utils/determineGiftToken.ts
```diff
@@ -0,0 +1,62 @@
...
+      new providers.JsonRpcProvider({
+        url: "https://nearrpc.aurora.dev",
+      })
+    )
+
+    // Currently there will be only one token with balance, but in future we may support multiple tokens in a gift
+    const tokenDiff = Object.fromEntries(
+      Object.entries(balances).filter(([_, balance]) => balance > 0n)
+    )
+    const token_ = tokenList.find((token) =>
```
cawabunga-bytes: Taking only a single token won't work for every scenarios. Say user is about to gift USDC. They have 100 USDC@Ethereum  and 100 USDC@Arbitrum. They want to gift $150, so gift will include two different tokens.

cawabunga-bytes (changes requested):

src/features/gift/utils/generateEscrowCredentials.ts
jobotics: Well, we already had this package and I found its simple and sufficient for this case. And as far as I know, it is used in near-api-js)

src/features/gift/components/GiftMakerForm.tsx
dr-block: no no, I meant just move out this code from this component, didn't refer to resuability, but just making this component smaller in terms of code size

src/features/gift/components/GiftMakerReadyDialog.tsx
dr-block: so they must stay in the same file , if they can not work separately ? 
I mean in that case we would have a very big app.js file and put all code in one place ...

src/features/gift/components/GiftTakerWidget.tsx
dr-block: yeah but creating separate file with constants would answer to my question forever, not only within this PR

src/features/gift/utils/parseGiftSecret.ts
jobotics: An error can be thrown at `normalizeNEP413Key` when we do `parseEscrowCredentials` [here](https://github.com/defuse-protocol/defuse-sdk/blob/feat/gift/src/features/gift/utils/generateEscrowCredentials.ts#L36)

src/features/gift/actors/giftMakerPublishingActor.ts
```diff
@@ -0,0 +1,177 @@
...
+      on: {
+        ACK_PUBLISHING_IMPOSSIBLE: "uncancellable",
+      },
+    },
+
+    publishing: {
+      invoke: {
+        src: "publishActor",
+        input: ({ context, event }) => {
+          return context?.multiPayload ?? event.input?.multiPayload
```
cawabunga-bytes: 1. Entire expression may return undefined
2. Could you explain why multipayload can be passed from two places (event and context)?

src/features/gift/actors/giftMakerPublishingActor.ts
```diff
@@ -0,0 +1,177 @@
...
+              intentHashes: ({ event }) => {
+                assert(event.output.tag === "ok")
+                return event.output.value
+              },
+            }),
+          },
+          {
+            target: "#(machine).idleUncancellable",
+            guard: {
+              type: "isNonceUsedError",
```
cawabunga-bytes: I know how nonce error may occur:
1. User publishes an intent
2. Relay receives request, executes a transaction and replies to the user
3. Network fails, user doesn't receive the response
4. User retries publishing after a while (1 hour?)
5. Relay returns simulation error with nonce used error

Are you sure it's a scenario that worth to handle explicitly? Am I missing something?

src/features/gift/actors/giftMakerPublishingActor.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    published: {
+      type: "final",
+      output: ({ context }) => {
+        assert(context.intentHashes, "intentHashes is not defined")
+        return {
+          giftStatus: "published",
+          intentHashes: context.intentHashes,
+        } satisfies GiftMakerPublishingActorOutput
+      },
```
cawabunga-bytes: You already have a data in context, so it's better to assemble the output in output function on the machine level.

src/features/gift/actors/giftMakerReadyActor.ts
```diff
@@ -0,0 +1,116 @@
...
+                event.output.giftStatus === "already_claimed_or_executed",
+            },
+          },
+          {
+            target: "idle",
+            actions: [
+              {
+                type: "logError",
+                params: {
+                  error: { reason: "EXCEPTION" },
```
cawabunga-bytes: It's not exception. `onError` is is the only who can produce such error. Please add correct errors here, and add missing onError callback if it is needed of course.

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+    complete: ({ self }, output: GiftMakerSignActorOutput) => {
+      self.send({ type: "COMPLETE", output })
+    },
+  },
+}).createMachine({
+  /** @xstate-layout N4IgpgJg5mDOIC5QAoC2BDAxgCwJYDswBKAOllynwKgGIIB7QkggN3oGswyL8AlMAGYBtAAwBdRKAAO9cgBdcjSSAAeiAIwBOACwkRADk36ArIZEA2Y+pFGANCACeicyQDsr9Sc0XjAZm3mNgBMAL4h9mhYeISk5JTUNGAATkn0SSRSADbocgJpqNyU-MLiyjLyivjKaggAtMZBJL6uxiK++tqaQa3mHeb2Tggu7p7G3pb+gV1hERg4BMSFVPi0AMIA8gCyAAoAMgCiACr7ohJIIOW4CkrnNdoiriTm6kFB5r6+xla+mv2OiLVzI99J8gq4ROogQZ9OpXGFwiB8PQIHBlJF5jEyrIrpVqgDrCImi02h0uj0TAMAUFISRtOp6a8Gi9-O0ZiB0dFFnFllAsRUbqA7oSwZpfD4Ya5tGNjH9BrUXvoSJp6S9lWLtPpXB02RyFqRMPRUFkwHJIHycQLVIgfroglKHi8ZR9PpSEPdaZLNK5la11L4Xq5zPCQkA */
+  initial: "signing",
+
+  context: ({ input }) => {
+    const nonce = randomDefuseNonce()
```
cawabunga-bytes: Nonce is auto-generated, so you can omit it.

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      /**
+       * If user has insufficient balance, we will generate a message with the full amount,
+       * and let the user know that they have insufficient balance.
+       */
+      const token = getAnyBaseTokenInfo(input.parsed.token)
+      tokenDiff = {
+        [token.defuseAssetId]: adjustDecimals(
+          // We need to negate the amount, as the balance is being reduced
```
cawabunga-bytes: Irrelevant comment

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+          input.parsed.amount.amount,
+          input.parsed.amount.decimals,
+          token.decimals
+        ),
+      }
+    }
+
+    const walletMessage = createTransferMessage(
+      [...Object.entries(tokenDiff)],
```
cawabunga-bytes: nit: just `Object.entries(tokenDiff)` is enough

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+      {
+        signerId: input.signerCredentials,
+        nonce: nonce,
+        referral: input.referral,
+        memo:
+          input.parsed.message.length > 0
+            ? input.parsed.message
+            : "Enjoy your gift!",
```
cawabunga-bytes: Let's not provide the default message. We can render a message on FE side.

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -0,0 +1,267 @@
...
+                </TokenAmountInputCard.DisplayPrice>
+              }
+            />
+          </div>
+          <div className="w-full mt-4">
+            <GiftMessageInput
+              inputSlot={
+                <GiftMessageInput.Input
+                  id="gift-message"
+                  name="message"
```
cawabunga-bytes, line 242: We need to add a note that this message is publicly readable (anyone can read this message onchain)

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  states: {
+    signing: {
+      invoke: {
+        id: "signRef",
+        src: "signActor",
+
+        input: ({ event, context }) => {
+          assertEvent(event, "xstate.init")
+          const input = event.input as GiftMakerSignActorInput
```
cawabunga-bytes: Can you double check if you need typecasting here?

src/features/gift/actors/giftMakerSignActor.ts
```diff
@@ -0,0 +1,231 @@
...
+          actions: [
+            { type: "logError", params: ({ event }) => event },
+            {
+              type: "complete",
+              params: { tag: "err", value: { reason: "EXCEPTION" } },
+            },
+          ],
+        },
+
+        // @ts-expect-error wtf???
```
cawabunga-bytes: So you just copy-pasted it? :D

dr-block (approved):

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -0,0 +1,333 @@
...
+            guard: ({ event }) => {
+              return event.output.giftStatus === "published"
+            },
+            target: "signed",
+          },
+          {
+            target: "editing",
+            actions: {
+              type: "setError",
+              params: { tag: "err", value: { reason: "EXCEPTION" } },
```
cawabunga-bytes: The same as a comment above, it's not exception, it should have a correct error.

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -0,0 +1,333 @@
...
+            },
+          },
+        ],
+        onError: {
+          target: "editing",
+          actions: [{ type: "logError", params: ({ event }) => event }],
```
cawabunga-bytes: And this is exception

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -0,0 +1,333 @@
...
+          assert(signData, "signData is not defined")
+
+          const form = context.formRef.getSnapshot()
+          const parsedValuesSnapshot = form.context.parsedValues.getSnapshot()
+
+          const parsedValues = parsedValuesSnapshot.context as {
+            [K in keyof typeof parsedValuesSnapshot.context]: NonNullable<
+              (typeof parsedValuesSnapshot.context)[K]
+            >
+          }
```
cawabunga-bytes: Maybe you can come up with better solution? Mine was written in hurry.

src/features/gift/actors/giftMakerPublishingActor.ts
jobotics: Oh, you are right, this is should be just `return context.multiPayload`, I'll correct it.

src/features/gift/actors/giftMakerSignActor.ts
jobotics: :D I love this comment

src/features/gift/actors/giftMakerPublishingActor.ts
jobotics: Can you elaborate a bit, do you mean we can remove this and use next scope of execution with setError, avoid it explicit, am I get you?

src/features/gift/components/GiftMakerForm.tsx
jobotics: Information added below the form

src/features/gift/actors/giftMakerPublishingActor.ts
cawabunga-bytes: Yes, sort of

src/features/gift/actors/giftMakerPublishingActor.ts
```diff
@@ -0,0 +1,158 @@
...
+      logger.error(event.error)
+    },
+    setError: assign({
+      error: (_, error: GiftMakerPublishingActorErrors) => error,
+    }),
+    clearError: assign({ error: null }),
+  },
+  guards: {
+    isOk: (_, params: { tag: "ok" | "err" }) => params.tag === "ok",
+    isNonceUsedError: (
```
cawabunga-bytes: This is not used anywhere.

cawabunga-bytes (approved):
Good work! I think I could leave more comments, but it is already good enough to be merged.

PR #329, opened 2025-03-10
feat: add extended MultiPayload schema for validating object

cawabunga-bytes:
# Background

`MultiPayload` is an object which consist of intents and a signature. This object is directly sent to intents.near contract. It has a very specific format. It's very hard to correctly serialize the object in required format, because it has non-obvious data manipulation.

# Summary

Developers who try to integrate with us struggle to assemble correct JSON. This PR adds and exports a schema which can be used to test JSON against.


src/utils/verifyWalletSignature.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { WalletSignatureResult } from "../types/swap"
+import { parsePublicKey, verifyAuthenticatorAssertion } from "./webAuthn"
+
+export async function verifyWalletSignature(
+  signature: WalletSignatureResult,
+  userAddress: string
+) {
+  const signatureType = signature.type
+  switch (signatureType) {
+    case "NEP413":
```
dr-block, line 13: just thinking about having them in one place as a constants, like

`` 
SIGNATURE_TYPES = {
"NEP413": "NEP413",
"ERC191": "ERC191"
...
}
``

in case of reusability, not duplications , avoiding typos, avoiding missing some types ...

dr-block (changes requested):

src/utils/verifyWalletSignature.ts
cawabunga-bytes: It would make sense if `signatureType` is a just string, but it's already typed. So it's not possible to write incorrect value. Also we have `signatureType satisfies never` which ensures you don't forget to handle all cases.

Screenshot 2025-03-11 at 09 19 18

src/features/otcDesk/components/OtcMakerTrades.tsx
```diff
@@ -315,7 +315,7 @@ function useValidateTrade(tradeTerms: TradeTerms) {
     ],
     queryFn: () => {
       return getDepositedBalances(
-        tradeTerms.userId,
+        tradeTerms.userId as DefuseUserId,
```
jobotics, line 318: Can we use here `userAddressToDefuseUserId`?

src/features/otcDesk/components/OtcTakerWidget.tsx
```diff
@@ -192,7 +192,7 @@ function OtcTakerValidationOrder({
     ],
     queryFn: () => {
       return getDepositedBalances(
-        tradeTerms.makerUserId,
+        tradeTerms.makerUserId as DefuseUserId,
```
jobotics, line 195: The same question above

src/features/otcDesk/components/OtcMakerTrades.tsx
cawabunga-bytes: Not really, because `userId` is already a string, which represents user within the protocol, it came directly  from multipayload. I admit that typecasting is bad practice, but in this case it is fine.

dr-block (approved):

jobotics (approved):

PR #332, opened 2025-03-12
feat: add support for validating TEX addresses in Zcash

cawabunga-bytes:
PR adds support for new TEX type addresses (https://zips.z.cash/zip-0320.html#specification). Basically it's bech32m address with specific prefix and specific size.

### DIFF
--- a/src/utils/validateAddress.ts
+++ b/src/utils/validateAddress.ts
@@ -60,10 +60,9 @@ export function validateAddress(
  */
 function validateZcashAddress(address: string) {
   // Transparent address validation
-  // t1 for P2PKH addresses, t3 for P2SH addresses
-  const tAddrRegex = /^t[13][a-km-zA-HJ-NP-Z1-9]{33}$/
   if (address.startsWith("t1") || address.startsWith("t3")) {
-    return tAddrRegex.test(address)
+    // t1 for P2PKH addresses, t3 for P2SH addresses
+    return /^t[13][a-km-zA-HJ-NP-Z1-9]{33}$/.test(address)
   }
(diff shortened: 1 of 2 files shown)


src/utils/validateAddress.ts
```diff
@@ -43,10 +44,41 @@ export function validateAddress(
...
+/**
+ * Validates Zcash addresses
+ * Supports:
+ * - Transparent addresses (t1, t3)
+ * - TEX addresses (tex1)
+ */
+function validateZcashAddress(address: string) {
+  // Transparent address validation
+  // t1 for P2PKH addresses, t3 for P2SH addresses
+  const tAddrRegex = /^t[13][a-km-zA-HJ-NP-Z1-9]{33}$/
```
dr-block: this const is used only if below condition is met, so what if we put it within IF , not outside if ?

dr-block (approved):

src/utils/validateAddress.ts
```diff
@@ -43,10 +44,40 @@ export function validateAddress(
...
+
+  // TEX address validation
+  const expectedHrp = "tex"
+  if (address.startsWith(`${expectedHrp}1`)) {
+    try {
+      const decoded = bech32m.decodeToBytes(address)
+      if (decoded.prefix !== expectedHrp) {
+        return false
+      }
+      return decoded.bytes.length === 20
```
jobotics, line 76: I think we can simplify a bit to `return decoded.prefix === expectedHrp && decoded.bytes.length === 20`

jobotics (approved):

PR #336, opened 2025-03-18
feat: gift persistent storages

jobotics:
## Implement triple storage system for Gifts persistency

Key Features:
Triple Storage Fallback: Data is first attempted to be retrieved from `localStorage`. If unavailable, it falls back to `sessionStorage`, and finally to `indexedDBStorage`.


src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -303,6 +313,16 @@ export const giftMakerRootMachine = setup({
             message: parsedValues.message,
           }
 
+          assert(context.intentHashes, "intentHashes is not defined")
```
dr-block: are we sure this is a valid assertion ? comparing with "intentHashes is not defined" is a good idea? do not we have any MORE valid check point ?

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -0,0 +1,115 @@
...
+                updatedAt: Date.now(),
+                tokenDiff: stringifiedTokenDiff,
+              },
+            ],
+          },
+        }))
+      },
+
+      removeGift: (giftId: string, user) => {
+        const userId =
```
dr-block: how about having function (even within this file) and call it getUserId that would contain this logic, as it is duplciated

src/features/gift/stores/indexedDBStorage.ts
```diff
@@ -0,0 +1,65 @@
+const GIFT_DB_NAME = "intents_sdk.gift_maker_gifts"
```
dr-block: is this the same  as GIFT_STORAGE_NAME ?  if yes, let;s reuse this in the store

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -0,0 +1,115 @@
...
+    const indexedData = await indexedDBStorage.getItem(name)
+    const data = localData || sessionData || indexedData
+    return data ? JSON.parse(data) : null
+  },
+  setItem: async (
+    name: string,
+    value: Record<DefuseUserId, GiftMakerHistory[]>
+  ) => {
+    const stringValue = JSON.stringify(value)
+    localStorageHandler.setItem(name, stringValue)
```
dr-block, line 46: what is the reason to set in ALL 3 places ?

dr-block (approved):

src/features/gift/actors/giftMakerRootMachine.ts
jobotics: Agree, I'll update this assertion so it will be more aligned with the context. If you mean something else please correct me.

src/features/gift/stores/giftMakerHistory.ts
jobotics: We are trying to reduce the chance of losing tokens in case one storage mechanism fails, is cleared, or the user loses the gift link, which would result in the funds being locked in the escrow account. This is simply an additional safety layer that does not impact the app's usability in any way.

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -0,0 +1,112 @@
...
+type Store = State & Actions
+
+export const tripleStorage = {
+  getItem: async (name: string) => {
+    const localData = localStorageHandler.getItem(name)
+    const sessionData = sessionStorageHandler.getItem(name)
+    const indexedData = await indexedDBStorage.getItem(name)
+    const data = localData || sessionData || indexedData
+    return data ? JSON.parse(data) : null
+  },
```
cawabunga-bytes, line 40: You don't need to access all storages, if you already retrieved the data.

cawabunga-bytes (approved):
I can't properly review indexedDB part, because i'm not familiar with it.

jobotics:
> I can't properly review indexedDB part, because i'm not familiar with it.

I'll do an additional check on it at the next pr (https://github.com/defuse-protocol/defuse-sdk/pull/339) with persistent data and try to break it.

PR #339, opened 2025-03-20
feat: add gift history widget

jobotics:
<details>
<summary>Screens:</summary>
Screenshot 2025-03-20 at 10 21 43
Screenshot 2025-03-20 at 10 22 36
Screenshot 2025-03-20 at 10 22 56
</details>


src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export function GiftHistoryWidget({
+  generateLink,
+  userAddress,
+  userChainType,
+  tokenList,
+}: GiftHistoryWidgetProps) {
+  const signerCredentials: SignerCredentials | null = useMemo(() => {
+    return userAddress != null && userChainType != null
```
dr-block: what if we check them just as boolean ? 
`return userAddress  && userChainType  ? .... {rest the same}`

src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+          credentialType: userChainType,
+        }
+      : null
+  }, [userChainType, userAddress])
+
+  return (
+    <WidgetRoot>
+      <SwapWidgetProvider>
+        <TabProvider>
+          {signerCredentials != null && (
```
dr-block: same here

src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+  )
+}
+
+type GiftHistoryProps = {
+  signerCredentials: SignerCredentials
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[]
+  generateLink: (giftPayload: GiftPayload) => string
+}
+
+function GiftHistory({
```
dr-block: pls let's move to another file and export from there, import and reuse here

src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return s.gifts[userId]
+  })
+
+  const [giftInfos, setGiftInfos] = useState<GiftInfos | null>(null)
+  useEffect(() => {
+    if (gifts == null) {
+      return
+    }
+    parseGiftInfos(tokenList, gifts).then((giftsResult) => {
+      setGiftInfos(giftsResult.unwrap())
```
dr-block: what is the benefit of double state ?

src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const userId = userAddressToDefuseUserId(
+      signerCredentials.credential,
+      signerCredentials.credentialType
+    )
+    return s.gifts[userId]
+  })
+
+  const [giftInfos, setGiftInfos] = useState<GiftInfos | null>(null)
+  useEffect(() => {
+    if (gifts == null) {
```
dr-block: we are going to end app with always loading mode should not we setLoading(false) here as well ?

src/features/gift/components/shared/GiftHistoryTabs.tsx
```diff
@@ -0,0 +1,30 @@
+import { cn } from "src/utils/cn"
+import { useTabContext } from "../../providers/TabProvider"
+
+export function GiftHistoryTabs() {
+  const { activeTab, setActiveTab } = useTabContext()
+  return (
+    <div className="flex flex-row justify-start items-center">
+      <button
```
dr-block: let's create a data structure and map it here , like
`tabs = [{key: "pending", title: "Pending"}, {key: "history", title: "History"}]`

src/features/gift/components/GiftHistoryWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[]
+  generateLink: (giftPayload: GiftPayload) => string
+}
+
+function GiftHistory({
+  signerCredentials,
+  tokenList,
+  generateLink,
+}: GiftHistoryProps) {
+  const pendingTab = usePendingTab()
```
dr-block: instead of boolean let's get activeTab key, f.e. it would be history, and after we can delete `pendingTab &&` and `historyTab &&` and use only one activeTab && ..... tag={activeTab}

src/features/gift/components/shared/GiftMakerHistoryItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+                    <CheckIcon weight="bold" />
+                  ) : (
+                    <CopyIcon weight="bold" />
+                  )}
+                </div>
+              </IconButton>
+            )}
+          </Copy>
+        )}
+        {tag === "pending" && (
```
dr-block: as before one is tag === "pending" I think this one can be merged with the previous one

src/features/gift/providers/TabProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type TabContextType = {
+  activeTab: "pending" | "history"
+  setActiveTab: (tab: "pending" | "history") => void
+}
+
+const TabContext = createContext<TabContextType | undefined>(undefined)
+
+export const TabProvider = ({ children }: { children: ReactNode }) => {
+  const [activeTab, setActiveTab] = useState<"pending" | "history">("pending")
```
dr-block: let's have reusable TabType "pending" | "history"

src/features/gift/providers/TabProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const useTabContext = () => {
+  const context = useContext(TabContext)
+  if (!context) {
+    throw new Error("useTabContext must be used within a TabProvider")
+  }
+  return context
+}
+
+export const usePendingTab = () => {
```
dr-block: guess after suggestions , those would be totally removed

src/features/gift/utils/parseGiftInfos.ts
```diff
@@ -0,0 +1,52 @@
...
+        logger.error(new Error("error parsing gift info", { cause: err }))
+        return {
+          tag: "failed",
+          ...gift,
+        }
+      }
+    })
+  )
+  return Ok({
+    pending: giftInfos.filter((giftInfo) => giftInfo.tag === "pending"),
```
dr-block: NIT, what if we have a function, filterByTab(tagName) and reuse it ?

dr-block (changes requested):

src/features/gift/components/GiftHistoryWidget.tsx
jobotics: `loading` is used to render gift skeletons until the gift data is fetched, and eventually `giftInfos` is used to properly map data between the two tabs: Pending and History.

dr-block: yeah, but why we should set giftsResult.unwrap() to state again, is n't it IN state and we can just use it, seems you take from state and set to another local state again ...

src/features/gift/components/shared/GiftHistory.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  if (loading) {
+    return <GiftHistorySkeleton />
+  }
+
+  return (
+    <div className="widget-container flex flex-col gap-4 p-5">
+      <GiftHistoryTabs />
+      <GiftClaimActorProvider signerCredentials={signerCredentials}>
+        {activeTab === "pending" &&
```
dr-block: why do we need to check activeTab === "pending" and activeTab === "history" ? 
let's remove double usage of GiftMakerHistoryItem and have just one , whatever tab would be active it would render that one
`
 <GiftMakerHistoryItem
              tag={activeTab}
              key={giftInfo.giftId}
              giftInfo={giftInfo}
              generateLink={generateLink}
              signerCredentials={signerCredentials}
            />
`

src/features/gift/components/shared/GiftHistoryEmpty.tsx
```diff
@@ -0,0 +1,7 @@
+export function GiftHistoryEmpty({ tag }: { tag: "pending" | "history" }) {
```
dr-block: TabType?

src/features/gift/components/shared/GiftHistoryTabs.tsx
```diff
@@ -0,0 +1,33 @@
+import { cn } from "src/utils/cn"
+import { useTabContext } from "../../providers/TabProvider"
+
+interface TabConfig {
+  key: "pending" | "history"
```
dr-block: TabType?

src/features/gift/components/GiftHistoryWidget.tsx
jobotics: Oh I see, it's because we use `monads` with result containing the success value, and to get it we should unwrap it. The other question is it necessary of use here `monads` and where we should use it.

src/features/gift/components/shared/GiftHistory.tsx
jobotics: But, we still should operate with two different arrays: `giftInfos?.pending` or `giftInfos?.claimed` and GiftMakerHistoryItem consume only items so we can't avoid such check, or do I missed something?

dr-block (approved):

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -249,6 +249,12 @@ export function GiftMakerForm({
                 />
               }
             />
+            {formValues.message && (
+              <MessageHint
+                text="Message is publicly readable and permanently stored on the blockchain. Anyone can view this message."
```
cawabunga-bytes: Message won't be publicly available, because it's encoded into url

src/utils/messageFactory.ts
```diff
@@ -347,7 +347,8 @@ export function makeInnerTransferMessage({
         intent: "transfer",
         tokens,
         receiver_id: receiverId,
-        memo,
+        // We don't want to include empty memo in the message
+        ...(memo && { memo }),
```
cawabunga-bytes: We don't need to put any memo to the intent

src/features/gift/utils/parseMultiPayload.ts
```diff
@@ -20,12 +20,19 @@ export function parseMultiPayloadTransferMessage(
...
     return null
   }
-  if (result.output.standard !== "nep413") {
-    logger.error(result.issues)
-    return null
+  const standard = result.output.standard
+  switch (standard) {
+    case "nep413":
+      return result.output.payload.message
+        .intents[0] as unknown as TransferIntentSubset
```
cawabunga-bytes: How do you know it exists?

cawabunga-bytes (approved):

PR #342, opened 2025-03-21
feat: add Account widget

cawabunga-bytes:
<details><summary>Screenshots</summary>
<p>

Screenshot 2025-03-21 at 15 18 22
Screenshot 2025-03-21 at 15 18 35
Screenshot 2025-03-21 at 15 18 53

</p>
</details>


src/hooks/useTokensUsdPrices.ts
jobotics, line 1: In the future, we might have several shared query hooks, so I think we might group them and keep them organized in a centralized location like `src/hooks/queries/`

cawabunga-bytes: Yeah, that might be a case, or we need to split them by features. Like balances in this PR.

src/features/account/queries/balanceQueries.ts
```diff
@@ -0,0 +1,60 @@
...
+    [string, { userId: null | DefuseUserId; tokenIds: string[] }]
+  >
+}
+
+export function createTransitBalanceQueryOptions({
+  userId,
+  tokenIds,
+}: { userId: null | DefuseUserId; tokenIds: string[] }) {
+  return {
+    queryKey: ["intents_sdk.transit_balance", { userId, tokenIds }],
```
jobotics, line 46: I would prefer to see `queryKey` assigned to the constant above for two reasons, first, it gives more readability to what key is used and second, it might be exported for cache invalidation.

src/features/account/components/AccountWidget.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function AccountWidget({
+  tokenList,
+  userAddress,
+  userChainType,
+  renderHostAppLink,
+}: AccountWidgetProps) {
+  const userId =
+    userAddress != null && userChainType != null
```
dr-block, line 27: what is the purpose != null ? we always do like this but I think just simply having userAddress && userChainType would be enough ... what do you think ?
btw, I saw a lot in the code that we check it before sending to userAddressToDefuseUserId, may be userAddressToDefuseUserId can expect nullable parameters and return null and we would not duplication always checking userAddress and userChainType ?

src/features/account/components/HoldingsIsland.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function Content({
+  isLoggedIn,
+  holdings,
+}: { isLoggedIn: boolean; holdings: Holding[] | undefined }) {
+  if (holdings?.length === 0 || !isLoggedIn) {
+    return <EmptyScreen />
+  }
+
+  if (holdings == null) {
```
dr-block, line 28: i think if you check this before EmptyScreen then you can change holdings?.length === 0 to holdings.length

src/features/account/components/shared/FormattedCurrency.tsx
```diff
@@ -0,0 +1,45 @@
...
+  mainPartClassName?: string
+  centsClassName?: string
+}) {
+  const formatter = new Intl.NumberFormat(locale, {
+    ...defaultFormatOptions,
+    ...formatOptions,
+  })
+  const parts = formatter.formatToParts(value)
+
+  const currencySymbol =
```
dr-block: I think you can simply map here 
[{type: {TYPE}, defaultValue: "DEFAULT"}], and whatever map wuld return  , you can map below in html, of course map can be memoable as well

src/features/account/components/shared/HoldingItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <div className="flex-1 flex flex-col items-start gap-0.5">
+        <div className="text-sm font-bold">{token.name}</div>
+        <div className="text-sm font-medium text-gray-11">{token.symbol}</div>
+      </div>
+
+      <div className="flex flex-col items-end gap-0.5">
+        <div className="text-sm font-medium" title={formatted}>
+          {shortFormatted ?? "-"} {token.symbol}
+        </div>
+        {usdValue != null ? (
```
dr-block, line 38: usdValue != null can be shortened to usdValue

src/features/account/components/shared/NavButton.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button } from "@radix-ui/themes"
+import type { ReactNode } from "react"
+import type {
+  HostAppRoute,
+  RenderHostAppLink,
+} from "../../../../types/hostAppLink"
+import { cn } from "../../../../utils/cn"
+
+type NavButtonProps = {
+  className?: string
```
dr-block: let's have not required items below than required once

src/features/account/queries/balanceQueries.ts
```diff
@@ -0,0 +1,60 @@
...
+import { assert } from "../../../utils/assert"
+import type { DefuseUserId } from "../../../utils/defuse"
+import type { BalanceMapping } from "../../machines/depositedBalanceMachine"
+
+export function createDepositedBalanceQueryOptions({
+  userId,
+  tokenIds,
+}: { userId: null | DefuseUserId; tokenIds: string[] }) {
+  return {
+    queryKey: ["intents_sdk.deposited_balance", { userId, tokenIds }],
```
dr-block, line 16: can we collect all keys in one place that we can have generalised string structure

src/features/account/components/HoldingsIsland.tsx
cawabunga-bytes: Not really. I render empty screen for unauthenticated user

src/features/account/components/AccountWidget.tsx
cawabunga-bytes: That's old problem with the interface, when we didn't have chainType, it's not even Chain Type. We should not touch this function.

cawabunga-bytes: Overall we should convert to defuse user id once and use it in every other parts.

src/features/account/components/shared/FormattedCurrency.tsx
cawabunga-bytes: Sorry I didn't get

src/features/account/components/shared/NavButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+  routeName,
+  renderHostAppLink,
+}: NavButtonProps) {
+  const children = (
+    <>
+      <Button
+        variant={variant === "primary" ? "solid" : "soft"}
+        color={variant === "primary" ? undefined : "gray"}
+        size="4"
+        className="w-full"
```
jobotics, line 32: As this is a link you should add `cursor-pointer` to class.

src/features/account/components/shared/HoldingItem.tsx
cawabunga-bytes: I prefer to use explicit null comparison, because 0 or "" are also falsy values, I don't want to  rely on type.

src/features/account/queries/balanceQueries.ts
cawabunga-bytes: I did something similar in other projects, but I didn't get any benefits. Just search for "queryKey" and you find all keys. You don't even need keys for invalidation. 

If you need to invalidate something, then invalidate everything. It's more reliable, because you may miss some cached keys. Consider any mutation as a full page refresh.

src/features/account/components/shared/NavButton.tsx
cawabunga-bytes: I didn't get this

src/features/account/components/AccountWidget.tsx
dr-block: can we at least instead have userAddress && userChainType && ... ??

src/features/account/components/shared/NavButton.tsx
cawabunga-bytes: Button should have needed style

src/features/account/components/AccountWidget.tsx
cawabunga-bytes: We should fix the source problem: user info interface is not correct. As soon it's fixed then everything should be fine.

src/features/account/components/shared/FormattedCurrency.tsx
dr-block: `const partList = [{type: "currency", defaultValue: "$"} .... ].map(data => parts.find((part) => part.type === data.type)?.value || data.defaultValue)`

`partlist.map(partItem => <span className={partItem === "fraction"  ? centsClassName : mainPartClassName}>{{partItem}}</span>)`

src/features/account/components/shared/NavButton.tsx
dr-block: I am talking about the order of props, when non required props are preferable to be below required props in the order they appear in the type

src/features/account/queries/balanceQueries.ts
dr-block: hmmm, i think it has 1 benefit
- first you combine all keys in one place - and you always can add a new key looking and structured like others, instead of searching and checking ALL queries structure ..

 invalidation is other discussion
marking it as a NIT, you can resolve if you prefer this approach

src/features/account/components/shared/NavButton.tsx
jobotics: Yeap, but for some reason she doesn't have them

src/features/account/components/shared/NavButton.tsx
cawabunga-bytes: I don't care about ordering btw. If you have preferences then we need to create lint rules. Otherwise "conventions" won't be followed.

src/features/account/queries/balanceQueries.ts
cawabunga-bytes: It's not that hard to find an example. I really don't think we need a separate file with keys. Let's create it if we have many keys and we start mixing them up.

src/features/account/components/shared/IntentsIcon.tsx
jobotics, line 1: Can it be located on FE side?

src/features/account/components/SummaryIsland.tsx
```diff
@@ -0,0 +1,96 @@
...
+            icon={<Gift weight="bold" className="size-5" />}
+          />
+        </div>
+      ) : (
+        renderHostAppLink(
+          "sign-in",
+          <ButtonCustom type="button" size="lg" className="w-full">
+            Sign in
+          </ButtonCustom>,
+          {}
```
jobotics, line 91: Is it necessary to have an empty object here?

src/features/account/components/shared/IntentsIcon.tsx
cawabunga-bytes: No, sdk should not rely on fe assets at all. Let's say this widget is rendered by 3rd party frontend, then where these assets come from?

src/features/account/components/shared/FormattedCurrency.tsx
cawabunga-bytes: I don't think that will be better.  Although this comment pointed me that default $ sign is incorrect. Like who said that it's dollars

src/features/account/components/SummaryIsland.tsx
cawabunga-bytes: Sort of, otherwise the function needs to handle null value, but it will be inconvenient for the host app. So it's easier to just pass empty object.

src/features/account/components/shared/FormattedCurrency.tsx
cawabunga-bytes: I removed default values (except "00" which is needed)

src/features/account/components/shared/NavButton.tsx
cawabunga-bytes: I reorder props

src/features/account/queries/balanceQueries.ts
cawabunga-bytes: I left a comment regarding this bellow:


> I did something similar in other projects, but I didn't get any benefits. Just search for "queryKey" and you find all keys. You don't even need keys for invalidation.
> 
> If you need to invalidate something, then invalidate everything. It's more reliable, because you may miss some cached keys. Consider any mutation as a full page refresh.
> ... 
> It's not that hard to find an example. I really don't think we need a separate file with keys. Let's create it if we have many keys and we start mixing them up.

jobotics (approved):

PR #343, opened 2025-03-21
chore: update gift history tabs

jobotics:
Changes: 
- Add removal Gift from history
- Add `Show more` Gifts

### DIFF
--- a/src/features/gift/actors/shared/getGiftInfo.ts
+++ b/src/features/gift/actors/shared/getGiftInfo.ts
@@ -24,6 +24,7 @@ export type GiftOpenSecretActorOutput =
     }
 
 export type GiftInfo = {
+  giftId?: string
   tokenDiff: Record<BaseTokenInfo["defuseAssetId"], bigint>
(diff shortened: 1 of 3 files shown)


src/features/gift/actors/shared/getGiftInfo.ts
```diff
@@ -24,6 +24,7 @@ export type GiftOpenSecretActorOutput =
     }
 
 export type GiftInfo = {
+  giftId?: string
```
cawabunga-bytes, line 27: I thought giftId has every gift?

cawabunga-bytes (approved):

dr-block (approved):
no comments except what @cawabunga-bytes mentioned

src/features/gift/actors/shared/getGiftInfo.ts
jobotics: For taker, we don't provide it as he doesn't have tabs history, so no point in having it

PR #345, opened 2025-03-24
fix: gift flow corrections

jobotics:


src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -150,8 +190,7 @@ export const giftMakerRootMachine = setup({
       id: "depositedBalanceRef",
       input: {
         tokenList: input.tokenList,
-        // `depositedBalanceActor` is any, so we explicitly safeguard it with `satisfies`
-      } satisfies InputFrom<typeof depositedBalanceMachine>,
+      },
```
cawabunga-bytes: You don't need to remove it. There is a comment that explains why it's required.

src/features/gift/components/shared/GiftMakerHistoryItem.tsx
```diff
@@ -29,9 +30,11 @@ export function GiftMakerHistoryItem({
   itemType: TabType
   signerCredentials: SignerCredentials
 }) {
+  const convertedTokenDiff = convertTokenDiffToBigInt(giftInfo.tokenDiff)
```
cawabunga-bytes: If I understand it correctly, you try to "fix" data type that came from local storage (which stores everything in strings). This component should accept correct data type, so you should correctly deserialize data somewhere else (probably in place where you get the data).

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -270,12 +277,31 @@ export function GiftMakerForm({
...
+    return false
+  }
+  if (formAmount.length === 0) {
+    return false
+  }
+  const formAmountBigInt = BigInt(
+    Math.round(Number.parseFloat(formAmount) * 10 ** tokenBalance.decimals)
+  )
+  return formAmountBigInt > tokenBalance.amount
+}
```
cawabunga-bytes: Could explain why you chose to have this logic here instead of machine? You already have `isValid` logic, maybe extend it?

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -357,11 +360,13 @@ export const giftMakerRootMachine = setup({
               amount: parsedValues.amount,
               message: parsedValues.message,
             },
+            depositedBalanceRef: context.depositedBalanceRef,
```
cawabunga-bytes: Why readyGift actor needs reference to depositedBalance machine?

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -131,6 +137,23 @@ export const giftMakerRootMachine = setup({
...
       self.send({ type: "COMPLETE_SIGN", params: event })
     },
+    addGiftToHistory: ({ context }) => {
+      assert(context.intentHashes, "intentHashes is not defined")
+      assert(context.signData, "signData is not defined")
+      const giftInfo = assambleReadyGiftInfo(context)
+      giftMakerHistoryStore.getState().addGift(
+        {
+          giftId: giftInfo.giftId,
+          intentHashes: context.intentHashes,
```
cawabunga-bytes: Shouldn't `intentHashes` be in `giftInfo`?

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -131,6 +137,23 @@ export const giftMakerRootMachine = setup({
     completeSign: ({ self }, event: GiftSignedResult) => {
       self.send({ type: "COMPLETE_SIGN", params: event })
     },
+    addGiftToHistory: ({ context }) => {
+      assert(context.intentHashes, "intentHashes is not defined")
+      assert(context.signData, "signData is not defined")
+      const giftInfo = assambleReadyGiftInfo(context)
+      giftMakerHistoryStore.getState().addGift(
```
cawabunga-bytes, line 148: I have concerns about `addGift`. It saves data to storages, some of them asynchronously, but you don't wait until successfully saved. Could you double check it?

src/features/gift/components/GiftTakerWidget.tsx
```diff
@@ -60,27 +64,41 @@ function GiftTakerScreens({
...
+  const intentStatus = useQuery({
+    queryKey: ["intents_status", intentHashes],
+    queryFn: async ({ signal }) => {
+      assert(intentHashes != null)
+      const intentHash = intentHashes[0]
+      assert(intentHash != null)
+      return waitForIntentSettlement(signal, intentHash)
+    },
+    enabled: intentHashes != null,
+  })
```
cawabunga-bytes: Could you explain why you put it here? Shouldn't it be in machine?

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -254,14 +281,46 @@ export function GiftMakerForm({
...
-            ? "Create gift link"
-            : "Confirm transaction in your wallet..."}
+          {renderButtonText(balanceInsufficient, editing, processing)}
         </ButtonCustom>
       </form>
     </div>
   )
 }
+
+function renderButtonText(
```
dr-block: how about renaming it to getButtonText ? as well as having this in separate file, may be utils ? under this domain ?
same for checkInsufficientBalance , into that utils file.

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -143,8 +143,13 @@ export const giftMakerRootMachine = setup({
...
       giftMakerHistoryStore.getState().addGift(
         {
-          ...giftInfo,
+          giftId: giftInfo.giftId,
           intentHashes: context.intentHashes,
+          tokenDiff: giftInfo.tokenDiff,
+          token: giftInfo.token,
+          secretKey: giftInfo.secretKey,
+          accountId: giftInfo.accountId,
+          message: giftInfo.message,
```
cawabunga-bytes: Why didn't you directly pass "giftInfo" object?

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -254,14 +281,46 @@ export function GiftMakerForm({
...
+  tokenBalance?: { amount: bigint; decimals: number }
+): boolean {
+  if (tokenBalance == null) {
+    return false
+  }
+  if (formAmount.length === 0) {
+    return false
+  }
+  const formAmountBigInt = BigInt(
+    Math.round(Number.parseFloat(formAmount) * 10 ** tokenBalance.decimals)
```
dr-block: can you explain why 10 ? may be we can name it ?

cawabunga-bytes (changes requested):

src/features/gift/components/GiftTakerForm.tsx
```diff
@@ -104,6 +109,9 @@ export function GiftTakerForm({
             : "Claim gift"
           : "Sign in to claim gift"}
       </ButtonCustom>
+      {intentHashes != null && intentHashes.length > 0 && (
```
dr-block: seems duplicated logic `intentHashes != null && intentHashes.length > 0` ...

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -373,3 +396,58 @@ export const giftMakerRootMachine = setup({
...
+type ReadyGiftInfo = GiftInfo & {
+  giftId: string
+  parsed: {
+    token: BaseTokenInfo | UnifiedTokenInfo
+    amount: TokenValue
+    message: string
+  }
+}
+
+function assambleReadyGiftInfo(
```
dr-block: i think those 3 functions can exist in different file. what do you think ?

src/features/gift/components/GiftMakerForm.tsx
```diff
@@ -100,6 +100,20 @@ export function GiftMakerForm({
       | ActorRefFrom<typeof giftMakerReadyActor>,
   }))
 
+  useEffect(() => {
```
dr-block: looks can be a separate hook like useCheckSignerCredentials(rootActorRef , signerCredentials) or {other name} ...

src/features/gift/components/GiftMakerForm.tsx
jobotics: Yeah, I thought about this, I decided not to mix the form logic's field elements with the balance and to keep it more pure and make it close to how we deal with it on deposit, swap and withdraw.

Moving there either requires extending the machine with balances context or shifting the check to the root machine. In both scenarios, it's just a guarded event, so we also need to extend the machine with some logic to get the `insufficient balance` state.

On the machine level, we have protection at publishing, so users still will not be able to create such gifts so I think we can move it to hook as this is mostly for button behavior.

src/features/gift/actors/giftMakerRootMachine.ts
jobotics: We can add it to assembles function but `intentHashes` important only for maker, so taker wont have it.

src/features/gift/actors/giftMakerRootMachine.ts
jobotics: Add additional failover mechanisms and test cases for `giftMakerHistoryStore`.

dr-block (approved):

src/features/gift/actors/giftMakerReadyActor.ts
```diff
@@ -36,7 +36,15 @@ export const giftMakerReadyActor = setup({
...
-    events: {} as { type: "FINISH" | "CANCEL_GIFT" },
+    events: {} as
+      | { type: "FINISH" | "CANCEL_GIFT" | "LOGOUT" }
+      | {
+          type: "LOGIN"
+          params: {
+            userAddress: string
+            userChainType: string
+          }
+        },
```
cawabunga-bytes, line 47: I don't get why ready actor needs login events.

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -102,6 +102,16 @@ export const giftMakerRootMachine = setup({
       void,
       GiftMakerReadyActorInput
     >,
+    settlingActor: fromPromise(
+      ({
+        input,
+        signal,
+      }: { input: { intentHashes: string[] }; signal: AbortSignal }) => {
+        const intentHash = input.intentHashes[0]
+        assert(intentHash, "intentHash is not defined")
```
cawabunga-bytes, line 111: you should provide correct data to the actor

src/features/gift/actors/giftMakerRootMachine.ts
```diff
@@ -126,16 +136,49 @@ export const giftMakerRootMachine = setup({
...
+      assert(context.signData, "signData is not defined")
+      const giftInfo = assembleGiftInfo(context)
+      giftMakerHistoryStore.getState().addGift(
+        {
+          ...giftInfo,
+        },
+        context.signData.signerCredentials
+      )
+    },
+    updateGiftToHistory: ({ context }) => {
```
cawabunga-bytes, line 155: nit: update to history?

cawabunga-bytes (approved):

src/features/gift/actors/giftMakerRootMachine.ts
jobotics: Yep, we can lose the escrow account after publishing if we accidentally refresh the page. To prevent this, we store it beforehand without the intent hash and then add it.

PR #346, opened 2025-03-24
feat: otc taker > support different decimals

dr-block:
Ticket: https://app.clickup.com/t/86err5192

### DIFF
--- a/src/features/otcDesk/hooks/useOtcTakerPreparation.ts
+++ b/src/features/otcDesk/hooks/useOtcTakerPreparation.ts
@@ -4,11 +4,19 @@ import { providers } from "near-api-js"
 import { logger } from "../../../logger"
 import { getDepositedBalances } from "../../../services/defuseBalanceService"
 import type { AggregatedQuote } from "../../../services/quoteService"
-import type { BaseTokenInfo, UnifiedTokenInfo } from "../../../types/base"
+import type {
+  BaseTokenInfo,
+  TokenValue,
+  UnifiedTokenInfo,
+} from "../../../types/base"
 import { assert } from "../../../utils/assert"
 import type { DefuseUserId } from "../../../utils/defuse"
+import { isBaseToken } from "../../../utils/token"
 import { getUnderlyingBaseTokenInfos } from "../../../utils/tokenUtils"
-import { fillWithMinimalExchanges } from "../utils/fillWithMinimalExchanges"
+import {
+  type TokenValues,
+  fillWithMinimalExchanges,
+} from "../utils/fillWithMinimalExchanges"
 import {
   type AggregatedQuoteErr,
   type QuoteExactInParams,
@@ -53,6 +61,21 @@ export function useOtcTakerPreparation({
         })
       )
 
+      const balancesWithTokenInfo = Object.keys(balances).reduce(
+        (acc: TokenValues, token) => {
+          acc[token] = {
+            amount: balances[token],
+            decimals: isBaseToken(tokenIn)
+              ? tokenIn.decimals
+              : tokenIn.groupedTokens.find((t) => t.defuseAssetId === token)
+                  ?.decimals,
+          } as TokenValue
+
+          return acc
+        },
+        {}
+      )
+
       logger.verbose("balances", { balances })
 
       const tokensToReceive: Record<string, bigint> = {}
@@ -69,7 +92,7 @@ export function useOtcTakerPreparation({
       logger.verbose("tokens breakdown", { tokensToReceive, tokensToSend })
 
       const fillResult = fillWithMinimalExchanges(
-        balances,
+        balancesWithTokenInfo,
         tokensToSend,
         BigInt(protocolFee)
       )
--- a/src/features/otcDesk/hooks/useOtcTakerPreparation.ts
+++ b/src/features/otcDesk/hooks/useOtcTakerPreparation.ts
@@ -4,11 +4,7 @@ import { providers } from "near-api-js"
 import { logger } from "../../../logger"
 import { getDepositedBalances } from "../../../services/defuseBalanceService"
 import type { AggregatedQuote } from "../../../services/quoteService"
-import type {
-  BaseTokenInfo,
-  TokenValue,
-  UnifiedTokenInfo,
-} from "../../../types/base"
+import type { BaseTokenInfo, UnifiedTokenInfo } from "../../../types/base"
 import { assert } from "../../../utils/assert"
 import type { DefuseUserId } from "../../../utils/defuse"
 import { isBaseToken } from "../../../utils/token"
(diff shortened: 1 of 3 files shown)


src/features/otcDesk/utils/fillWithMinimalExchanges.test.ts
```diff
@@ -285,25 +295,160 @@ describe("fillWithMinimalExchanges", () => {
...
+          toAmount: 50n,
+          fee: 0n,
+        },
+      ]),
+    })
+  })
+
+  it("multi chain can cover, different decimals (with 2 chains)", () => {
+    const balances: TokenValues = {
+      A: { amount: 10012345678n, decimals: 18 },
```
jobotics, line 339: Can we simplify variables in this test avoiding such big values 10012345678n? Same as in the test above as it blurs understanding.

dr-block: that has 2 reason 
- it is a big as if it is small then I should have 0.0000 something for 8 decimals, I kept 8 decimals some number 5n or 50n , and after this becomes automatically bigger  
- about 12345678.... this kind of style, I didn't want to have like clear 100000 numbers like this as to let Math working and doing some things in the background .... also I thought as previous tests are written with nice (0 containing numbers), then let's have this with some ugly numbers ...

what do you think ?

jobotics: Not sure, but we can try. One thing to consider is that the preferred pattern should be described in some way to show that it’s not just a random number, but a deliberate choice.

dr-block: the case here is that there is no patter, as there is no pattern what would user type ... 1234... and 9876... as unpaternized ... i tried to not have patterns , thinking as that there can be any number.

src/features/otcDesk/hooks/useOtcTakerPreparation.ts
```diff
@@ -53,6 +61,21 @@ export function useOtcTakerPreparation({
...
+            amount: balances[token],
+            decimals: isBaseToken(tokenIn)
+              ? tokenIn.decimals
+              : tokenIn.groupedTokens.find((t) => t.defuseAssetId === token)
+                  ?.decimals,
+          } as TokenValue
+
+          return acc
+        },
+        {}
```
cawabunga-bytes: It's better to type cast here. `{} as TokenValues`. Then acc will correctly typed.

src/features/otcDesk/hooks/useOtcTakerPreparation.ts
```diff
@@ -53,6 +61,21 @@ export function useOtcTakerPreparation({
...
       )
 
+      const balancesWithTokenInfo = Object.keys(balances).reduce(
+        (acc: TokenValues, token) => {
+          acc[token] = {
+            amount: balances[token],
+            decimals: isBaseToken(tokenIn)
+              ? tokenIn.decimals
+              : tokenIn.groupedTokens.find((t) => t.defuseAssetId === token)
+                  ?.decimals,
```
cawabunga-bytes: This may produce undefined. You can use `assert` to ensure it is non null.

jobotics (approved):

src/features/otcDesk/utils/fillWithMinimalExchanges.test.ts
```diff
@@ -285,25 +295,160 @@ describe("fillWithMinimalExchanges", () => {
...
+          fromAmount: 4n,
+          toAmount: 3n,
+          fee: 1n,
+        },
+      ],
+      success: true,
+    })
+  })
+
+  it("multi chain can cover, different decimals (with 3 chains)", () => {
```
cawabunga-bytes, line 377: Let's also add a case where amounts are deducted from 3 tokens like this:
```ts
    const balances: TokenValues = {
      A: { amount: 3n, decimals: 2 }, // cover 3A (without fee)
      B: { amount: 3n, decimals: 1 }, // cover 10A = 1B (with fee it will be 2B)
      C: { amount: 33n, decimals: 2 }, // cover 3A = 3C (with fee it will be 4C)
    }
    const required: TokenBalances = {
      A: 16n,
    }
```

It's funny because it will suboptimal for the user.

src/features/otcDesk/utils/fillWithMinimalExchanges.ts
```diff
@@ -96,12 +172,19 @@ function findOptimalTokenSources(
...
   const result: FillResult = {
-    remainingBalances: { ...balances },
+    remainingBalances: Object.keys(balancesWithTokenInfo).reduce(
+      (acc: TokenValues, curr: string) => {
+        acc[curr] = Object.assign({}, balancesWithTokenInfo[curr]) as TokenValue // like copying as we make actions with this object below
+
+        return acc
+      },
+      {}
+    ),
```
cawabunga-bytes: You can you structuredClone to copy objects

src/features/otcDesk/utils/fillWithMinimalExchanges.test.ts
```diff
@@ -285,25 +295,160 @@ describe("fillWithMinimalExchanges", () => {
...
+          toToken: "A",
+          fromAmount: 10012345678n,
+          toAmount: 10012345678n,
+          fee: 0n,
+        },
+        {
+          fromToken: "B",
+          toToken: "A",
+          fromAmount: 3n,
+          toAmount: 2n,
```
cawabunga-bytes: I think it should be in terms of token to. Then you can keep using checkTotalOut in the test.

src/features/otcDesk/utils/fillWithMinimalExchanges.test.ts
dr-block: think in exact this scenario it would cover with to as actually you have converted to this

```
const balances: TokenValues = {
      A: { amount: 3n, decimals: 2 }, 
      B: { amount: 30n, decimals: 2 }, 
      C: { amount: 33n, decimals: 2 }, 
    }
```

first it would take from A 3n, after 13n from B ... 
think we should ask required , something like 50n ...

src/features/otcDesk/utils/fillWithMinimalExchanges.test.ts
dr-block: didn't get your point, sorry

cawabunga-bytes: Here `toAmount: 2n,` is represented in token B (fromToken). 
It's better to have it in token A (toToken) `toAmount: 200...000n,`

dr-block: that was the question I asked today you .... should it work like that or like it is now coded .... 
I tried to sign like it is now, and it was working as expected ... will check with your logic as well

cawabunga-bytes (approved):
Great job!

PR #349, opened 2025-03-24
feat: swaps the network Select with a Modal that also shows unsupported chains

hyper54n3:
[Figma](https://www.figma.com/design/FGmbApkVo2Nb4NH259CAME/Defuse?node-id=4596-14812&t=liXz0HL8uTptHZRF-4)

- [X] Create a new modal for selecting the network
- [X] Swap to the modal in Deposit form
- [X] Swap to the modal in Withdraw form
- [x] add (?) tooltip to Unsupported Networks

Localhost preview:
image


src/components/Modal/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,143 @@
...
+              placeholder="Search"
+              query={searchValue}
+              setQuery={setSearchValue}
+            />
+          </div>
+        </div>
+
+        <div className="z-10 flex-1 overflow-y-auto  -mr-[var(--inset-padding-right)] pr-[var(--inset-padding-right)]">
+          {Object.keys({ ...filteredChains, ...disabledChains }).length ===
+          0 ? (
```
jobotics: You can utilize this in func as it is used in several places, something like `hasChains({ ...filteredChains, ...disabledChains })` and below the same `hasChains(filteredChains)`.

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -185,27 +198,29 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
...
-                    <Select.Hint>
-                      {Object.keys(chainOptions).length === 1
-                        ? "This network only"
-                        : "Network"}
-                    </Select.Hint>
-                  }
-                />
-              )}
+              render={({ field }) => {
+                const label =
```
jobotics: What is the point of placing those variables here? Why not move them out of JSX up or place them directly in the component below?

src/utils/blockchain.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Converts a BlockchainEnum value to its corresponding SupportedChainName
+ * @param blockchain The BlockchainEnum value to convert
+ * @returns The corresponding SupportedChainName
+ */
+export function blockchainToChainName(
+  blockchain: BlockchainEnum
+): SupportedChainName {
+  // Split the blockchain string by ':' and take the first part
+  const blockchainToChainMap: Record<BlockchainEnum, SupportedChainName> = {
```
jobotics: Btw, we already have such a map `networkSelectToLabel`, so maybe we can update it to?

src/features/deposit/components/DepositForm/index.tsx
hyper54n3: Just a bit more readable than straight in the props, but really a question of preference.
I'll update it to have them in the props directly

jobotics (approved):

src/utils/blockchain.ts
```diff
@@ -0,0 +1,52 @@
...
+    [BlockchainEnum.BITCOIN]: "bitcoin",
+    [BlockchainEnum.SOLANA]: "solana",
+    [BlockchainEnum.DOGECOIN]: "dogecoin",
+    [BlockchainEnum.TURBOCHAIN]: "turbochain",
+    [BlockchainEnum.AURORA]: "aurora",
+    [BlockchainEnum.XRPLEDGER]: "xrpledger",
+    [BlockchainEnum.ZCASH]: "zcash",
+    [BlockchainEnum.GNOSIS]: "gnosis",
+    [BlockchainEnum.BERACHAIN]: "berachain",
+  }
```
cawabunga-bytes: I think we already have such function in src/utils/adapters.ts:19

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -269,7 +280,12 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
   )
 }
 
-function getBlockchainsOptions(): Record<
+export function isAuroraVirtualChain(network: BlockchainEnum): boolean {
+  const virtualChains = [BlockchainEnum.TURBOCHAIN]
```
cawabunga-bytes: Aurora is also virtual chain, is it skipped deliberately? I think such info should be stored at some constant config.

Overall current chains configuration is big mess. Chain configuration is evenly distributed across entire codebase. For example, we have PoA Bridge chain id -> internal chain id mapping in one file; aurora engine contract ids in another file; chain labels and icons in 3rd file; evm chain id in 4th file and so on. 

So as for now it would better to reuse existing config (for example, use `siloToSiloAddress` or `auroraEngineContractId` to determine if it is an aurora powered chain). And as a long term it would nice to aggregate chain configs in one place.

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -388,28 +410,31 @@ export const WithdrawForm = ({
...
+                  onClick={() => openModalSelectNetwork()}
+                  hint={
+                    <Select.Hint>
+                      {Object.keys(blockchainSelectItems).length === 1
+                        ? "This network only"
+                        : "Network"}
+                    </Select.Hint>
+                  }
+                  disabled={
+                    blockchainSelectItems &&
```
cawabunga-bytes: nit: `blockchainSelectItems` is always defined

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -188,11 +195,26 @@ export const WithdrawForm = ({
...
 
-  const { setModalType, data: modalSelectAssetsData } = useModalController<{
+  const { data: modalSelectAssetsData } = useModalController<{
     modalType: ModalType
     token: BaseTokenInfo | UnifiedTokenInfo | undefined
   }>(ModalType.MODAL_SELECT_ASSETS)
 
+  const { setModalType, onCloseModal } = useModalStore((state) => state)
+
+  const onChangeNetwork = (network: BlockchainEnum) => {
```
cawabunga-bytes: I suggest not using `BlockchainEnum`, because it is specific to PoA bridge, but we operate with other "bridges" too. For example, Aurora Engine chains have own bridge (transfer via silo contract), Near has direct withdrawal (but still we can consider it a "direct" bridge). And Omni bridge is also coming. So I suggest using `SupportedChainName` instead.

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -188,11 +195,26 @@ export const WithdrawForm = ({
...
+    setValue("blockchain", blockchainToChainMap[network])
+    onCloseModal()
+  }
+
+  const openModalSelectNetwork = () => {
+    setModalType(ModalType.MODAL_SELECT_NETWORK, {
+      token,
+      selectNetwork: onChangeNetwork,
+      selectedNetwork: chainNameToBlockchain(blockchain),
+    })
```
cawabunga-bytes: Personally I don't like such dialogs/modals imperative control. I prefer a declarative approach, where you just render the modal and pass `isOpen` prop to it. 1. It is more conventional to React (and other declarative frameworks). 2. Current implementation hard to use and properly write types. But it's up to you to choose what the best here is.

rfdlp: I agree with you here. Did it this way to follow the existing pattern. Will change it

src/components/Modal/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,143 @@
...
+          label: chain.label,
+          icon: chain.icon,
+          value: chain.value,
+        }
+      }
+    })
+    return Object.fromEntries(
+      Object.entries(disabledChains).filter(
+        ([key, chain]) =>
+          chain.label.toLowerCase().includes(searchValue.toLowerCase()) ||
```
dr-block: seems a duplicated logic of filtration, can we make it reusable ?

src/components/Modal/NetworksList.tsx
```diff
@@ -0,0 +1,50 @@
+import { Text } from "@radix-ui/themes"
+import clsx from "clsx"
+import type { ReactNode } from "react"
+import { isAuroraVirtualChain } from "src/features/deposit/components/DepositForm"
+import type { BlockchainEnum } from "src/types/interfaces"
+import { PoweredByAuroraLabel } from "../PoweredByAuroraLabel"
+
+interface NetworkListProps {
+  networks: Record<string, { label: string; icon: ReactNode; value: string }>
```
dr-block: if chain value can be only BlockchainEnum, why we have here string and below BlockchainEnum ?

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -186,24 +199,22 @@ export const DepositForm = ({ chainType }: { chainType?: ChainType }) => {
...
-                  }}
-                  value={
-                    getDefaultBlockchainOptionValue(token) || network || ""
-                  }
-                  onChange={field.onChange}
-                  name={field.name}
+                <SelectTriggerLike
+                  label={chainOptions[network ?? ""]?.label ?? "Select network"}
+                  icon={chainOptions[network ?? ""]?.icon ?? <EmptyIcon />}
+                  onClick={() => openModalSelectNetwork()}
```
dr-block: onClick ={openModalSelectNetwork}

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -388,28 +410,31 @@ export const WithdrawForm = ({
...
+              render={({ field }) => (
+                <SelectTriggerLike
+                  label={
+                    blockchainSelectItems[field.value]?.label ??
+                    "Select network"
+                  }
+                  icon={
+                    blockchainSelectItems[field.value]?.icon ?? <EmptyIcon />
+                  }
+                  onClick={() => openModalSelectNetwork()}
```
dr-block: onClick= openModalSelectNetwork

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -388,28 +410,31 @@ export const WithdrawForm = ({
...
+                  hint={
+                    <Select.Hint>
+                      {Object.keys(blockchainSelectItems).length === 1
```
dr-block: why not reuse Object.keys(blockchainSelectItems).length === 1?

src/components/Modal/ModalNoResults.tsx
```diff
@@ -3,15 +3,17 @@ import type { FC } from "react"
...
   handleSearchClear: () => void
+  text?: string
 }
 
 export const ModalNoResults: FC<ModalNoResultsProps> = ({
   handleSearchClear,
+  text = "No assets found",
 }) => {
   return (
     <div className="flex justify-center items-center flex-col gap-2.5">
```
vashchukmaksim, line 14: **not related to this PR** but hm... we use props from Radix and classes from Tailwind notation together 🤔

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const availableChains = useMemo(() => availableChainsForToken(token), [token])
+  const filteredChains = filterChains(availableChains, searchValue)
+
+  const disabledChains = () => {
+    const disabledChains: Record<
+      string,
+      { label: string; icon: ReactNode; value: string }
+    > = {}
+    Object.values(chains).map((chain) => {
```
vashchukmaksim, line 48: We don't assign `map` result anywhere but it in fact creates a new array while what we need is a side effect to populate an object. I would change it to `forEach` instead or even better to `.reduce` and avoid assignment to empty object

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+      if (!filteredChains[chain.value]) {
+        disabledChains[chain.value] = chain
+      }
+    })
+    return filterChains(disabledChains, searchValue)
```
vashchukmaksim, line 53: nit: We kinda do another loop over the same chains again here, can we avoid that?

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const onChangeNetwork = (network: SupportedChainName) => {
+    selectNetwork(network)
+    onClose()
+  }
+
+  const availableNetworks = Object.keys(filteredChains).map(
+    (key) => key as BlockchainEnum
+  )
+  const disabledNetworks = Object.keys(disabledChains()).map(
```
vashchukmaksim, line 64: Maybe just wrap `disabledChains` is `useMemo` above rather than calling it here on every render

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+            <SearchBar
+              placeholder="Search"
+              query={searchValue}
+              setQuery={setSearchValue}
+            />
+          </div>
+        </div>
+
+        <div className="z-10 flex-1 overflow-y-auto  -mr-[var(--inset-padding-right)] pr-[var(--inset-padding-right)]">
+          {Object.keys({ ...filteredChains, ...disabledChains }).length ===
```
vashchukmaksim, line 90: Hm... but `disabledChains` is a function

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+          0 ? (
+            <ModalNoResults
+              text="No networks found"
+              handleSearchClear={() => setSearchValue("")}
+            />
+          ) : (
+            <div className="flex flex-col gap-2 divide-y divide-gray-300">
+              {Object.keys(filteredChains).length > 0 && (
```
vashchukmaksim, line 98: Since here we do this check for `NetworkList` and use `availableNetworks` that we've already got using `Object.keys(filteredChains)` above it seems logical check for it directly: `availableNetworks.length > 0`

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,140 @@
...
+                <div className="flex flex-col gap-2">
+                  <NetworkList
+                    networks={availableNetworks}
+                    selectedNetwork={selectedNetwork}
+                    onChangeNetwork={onChangeNetwork}
+                  />
+                </div>
+              )}
+              {Object.keys(disabledNetworks).length > 0 && (
```
vashchukmaksim, line 107: Same as for `availableNetworks` above

vashchukmaksim: ook, you actually do check `disabledNetworks` and not `disabledChains` but it's an array, why `Object.keys` then?

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -80,12 +86,18 @@ export const DepositForm = ({
...
-  const network = blockchain ? assetNetworkAdapter[blockchain] : null
-
   const { setModalType, payload, onCloseModal } = useModalStore(
     (state) => state
   )
 
+  const onCloseNetworkModal = () => setIsNetworkModalOpen(false)
+
+  const onChangeNetwork = (network: SupportedChainName) => {
+    // console.log("onChangeNetwork", network)
```
vashchukmaksim, line 96: leftover console.log

src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -195,25 +210,33 @@ export const DepositForm = ({
...
-                  }
-                />
+                <>
+                  <SelectTriggerLike
+                    label={chainOptions[networkEnum]?.label ?? "Select network"}
+                    icon={chainOptions[networkEnum]?.icon ?? <EmptyIcon />}
+                    onClick={() => setIsNetworkModalOpen(true)}
+                    hint={
+                      <Select.Hint>
+                        {Object.keys(chainOptions).length === 1
```
vashchukmaksim, line 220: nit: we iterate twice with `Object.keys` to do length check. Let's maybe move it outside of JSX: `const isOnlyChainOptions = Object.keys(chainOptions).length === 1`

src/utils/blockchain.ts
```diff
@@ -0,0 +1,24 @@
...
+  candidates: Record<string, { label: string; icon: ReactNode; value: string }>,
+  searchValue: string
+) => {
+  if (searchValue === "") {
+    return candidates
+  }
+  return Object.fromEntries(
+    Object.entries(candidates).filter(
+      ([key, chain]) =>
+        chain.label.toLowerCase().includes(searchValue.toLowerCase()) ||
```
vashchukmaksim, line 19: We do `searchValue.toLowerCase()` three times here (even though sometimes one or two won't be called). Let's do it once after `if` check: `const lowerCaseSearch = searchValue.toLowerCase()` and use it for checks

src/components/Modal/ModalNoResults.tsx
capybara-madness, line 14: not related to this PR but hm... we use props from Radix and classes from Tailwind notations together 🤔

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: We don't assign map result anywhere but it in fact creates a new array while what we need is a side effect to populate an object. I would change it to forEach instead or even better to .reduce and avoid assignment to empty object

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: nit: We kinda do another loop over the same chains again here, can we avoid that?

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: Maybe just wrap `disabledChains` in `useMemo` above rather than calling it here on every render

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: Hm... but disabledChains is a function, do not spread it 😇 Actually as for below we can just simplify a check and avoid spreading and creating a new object with it: `availableNetworks.length === 0 && disabledNetworks.length === 0`

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: Since here we do this check for `NetworkList` and use `availableNetworks` that we've already got using `Object.keys(filteredChains)` above it seems logical check for it directly: `availableNetworks.length > 0`

src/components/Network/ModalSelectNetwork.tsx
capybara-madness: ook, you actually do check `disabledNetworks` and not `disabledChains` but it's an array already, why `Object.keys` then?

src/features/deposit/components/DepositForm/index.tsx
capybara-madness: leftover `console.log`

src/features/deposit/components/DepositForm/index.tsx
capybara-madness: nit: we iterate twice with `Object.keys` to do length check. Let's maybe move it outside of JSX: `const isOnlyChainOption = Object.keys(chainOptions).length === 1`

src/utils/blockchain.ts
capybara-madness: We do `searchValue.toLowerCase()` three times here (even though sometimes one or two won't be called). Let's do it once after if check: `const lowerCaseSearch = searchValue.toLowerCase()` and use it for checks

src/components/Network/ModalSelectNetwork.tsx
```diff
@@ -0,0 +1,137 @@
...
+            <SearchBar
+              placeholder="Search"
+              query={searchValue}
+              setQuery={setSearchValue}
+            />
+          </div>
+        </div>
+
+        <div className="z-10 flex-1 overflow-y-auto  -mr-[var(--inset-padding-right)] pr-[var(--inset-padding-right)]">
+          {[...availableNetworks, ...disabledNetworks].length === 0 ? (
```
capybara-madness, line 88:
```suggestion
          {availableNetworks.length === 0 && disabledNetworks.length === 0 ? (
```

- not creating a new array
- constant time complexity

capybara-madness (approved):

cawabunga-bytes (approved):

PR #358, opened 2025-03-25
fix: multi decimal support for swap and otc maker ticket: 86ery8ctn

dr-block:


src/services/quoteService.ts
```diff
@@ -218,18 +219,64 @@ function assert(condition: unknown, msg?: string): asserts condition {
...
+      return 1
+    }
+    if (b.decimals > a.decimals) {
+      return -1
+    }
+
+    if (balances[a.defuseAssetId] && balances[b.defuseAssetId]) {
+      const maxDecimalBetweenAandB = Math.max(a.decimals, b.decimals) // taking max from decimals to ave cleaner comparing
+      const aBalanceAdjusted = adjustDecimals(
+        balances[a.defuseAssetId] as bigint,
```
cawabunga-bytes: Please don't you type casting, if you expect correct data then just use `assert`

src/services/quoteService.ts
```diff
@@ -218,18 +219,64 @@ function assert(condition: unknown, msg?: string): asserts condition {
...
+): BaseTokenInfo[] {
+  return structuredClone(uniqueTokensIn).sort((a, b) => {
+    if (b.decimals < a.decimals) {
+      return 1
+    }
+    if (b.decimals > a.decimals) {
+      return -1
+    }
+
+    if (balances[a.defuseAssetId] && balances[b.defuseAssetId]) {
```
cawabunga-bytes: It's not clear if you consciously handle `0` or not. That's why I use explicit comparison `balances[a.defuseAssetId] != null`.

src/services/quoteService.ts
dr-block: I will add assert on top of this, but sometimes assert is not enough, it keeps yelling about type undefined.

cawabunga-bytes: You can write like this
```
const var = foo[bar]
assert(var != null)
// var is non-nullable
```

or (in some case it's completely valid)

```
arr.map((item, i, list) => {
  // biome-ignore lint/style/noNonNullAssertion: `i` is always in bounds
  doSomething(list[i]!)
})
```

cawabunga-bytes (approved):

PR #360, opened 2025-03-26
fix: gift data serialization

jobotics:


src/features/gift/stores/giftMakerHistory.test.ts
```diff
@@ -61,3 +92,75 @@ describe("tripleStorage", () => {
...
+                "token": {
+                  "decimals": 6,
+                  "groupedTokens": [],
+                  "icon": "icon",
+                  "name": "USD Coin",
+                  "symbol": "USDC",
+                  "unifiedAssetId": "usdc",
+                },
+                "tokenDiff": {
+                  "nep141:usdc": "1000",
```
dr-block: do we have a test coverage for Big Int case, which was the bug we found?

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -36,109 +41,131 @@ type Actions = {
...
+async function executeStorageOperations<T>(
+  operations: StorageOperation<T>[]
+): Promise<(T | null)[]> {
+  return Promise.all(
+    operations.map(async ({ operation, storageName, operationType }) => {
+      try {
+        return await operation()
+      } catch (error) {
+        const action =
+          operationType === "fetch"
```
dr-block: it is a bit hard to read this drill down ... may be switch ?

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -36,109 +41,131 @@ type Actions = {
...
-        new Error("Failed to update data in IndexedDB", { cause: error })
-      )
-    }
+
+  updateItem: async (name: string, value: GiftData) => {
+    const stringValue = JSON.stringify(serializeGiftData(value))
+    await executeStorageOperations([
+      {
+        operation: () => localStorageHandler.setItem(name, stringValue),
+        storageName: "localStorage",
```
dr-block, line 152: the only need for  storageName and  operationType as far as I can see is just for catch/logger ... 
what if we just remove those, and in the catch we log the operation ... would not that be enough for debug process?

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -36,109 +41,131 @@ type Actions = {
...
-    sessionStorageHandler.setItem(name, stringValue)
-    try {
-      await indexedDBStorage.setItem(name, stringValue)
-    } catch (error) {
-      logger.error(
-        new Error("Failed to set data in IndexedDB", { cause: error })
-      )
-    }
+
+  setItem: async (name: string, value: GiftData) => {
```
dr-block: as we are now batch setting item ... this means that there can be a case that we duplicate something in 3 places .... did this what we wanna ? 

what I was thinking is that the setter/updater/remover should run ONE storage ... getter should take from 3 of them and merge

by this we would be secure of duplications ....

otherwsie somewhere else we should have code to merge by unique ID. am I right ?

src/features/gift/stores/giftMakerHistory.test.ts
jobotics: Yes, this is the test that covers it.

src/features/gift/stores/giftMakerHistory.ts
jobotics: Well, this will reduce our observability by three times as we will not know what system cause the problem.

src/features/gift/stores/giftMakerHistory.ts
jobotics: No, first—we use separate records individually and do not merge storages. Second, all storages are mirrors of each other, so we don't need to merge them, and we are satisfied with retrieving at least one.

dr-block (approved):

src/features/gift/stores/giftMakerHistory.ts
```diff
@@ -36,109 +45,144 @@ type Actions = {
...
+          case "remove":
+            action = "remove data from"
+            break
+          default:
+            action = "interact with"
+        }
+        logger.error(
+          new Error(`Failed to ${action} ${storageName}`, { cause: error })
+        )
+        return null
```
cawabunga-bytes, line 82: You decided ignore error and then just override with new data?

jobotics: You are right, this is important. I think it's better to do it in separate PR.

PR #361, opened 2025-03-26
fix: remove default Chrome focus-visible blue outline on inputs

jobotics:


src/components/Form/FieldComboInput.tsx
```diff
@@ -125,7 +125,7 @@ export const FieldComboInput = <T extends FieldValues>({
             disabled={disabled}
             autoComplete="off"
             className={clsx(
-              "bg-gray-50 w-full text-3xl font-medium placeholder-black border-transparent focus:border-transparent focus:ring-0 dark:bg-black-900 dark:placeholder-white px-0",
+              "bg-gray-50 w-full text-3xl font-medium placeholder-black border-transparent focus:border-transparent focus:ring-0 dark:bg-black-900 dark:placeholder-white px-0 outline-none",
```
dr-block, line 128: i can see here a duplication ... it seems we wanna remove some default styles , but end up having a bug on chrome ...
what was the purpose ? remove input border ?
how about having custom no-input-border TW classname that would collect some of TW classnames within

f.e. no-input-border: "border-0 bg-transparent outline-none .... {I am not sure exact which once should be used ... but seems you know which of them are reliable ....}"

what do you think ?

jobotics: I agree, but this is mostly a hotfix rather than refactoring. The aimed was to remove this border.

Screenshot 2025-03-26 at 09 51 14

dr-block (approved):

cawabunga-bytes (approved):

src/components/Form/FieldComboInput.tsx
cawabunga-bytes: It should be fixed on tailwind side rather then explicit reset on each affected element. Somewhere in main.css in 
```
@layer base {
...
```

PR #365, opened 2025-03-27
feat(withdraw): add transaction on destination chain

cawabunga-bytes:
<details><summary>Screenshots</summary>
<p>

Screenshot 2025-03-27 at 13 31 57
Screenshot 2025-03-27 at 13 31 34

</p>
</details>

### DIFF
--- a/src/features/withdraw/components/WithdrawWidget.tsx
+++ b/src/features/withdraw/components/WithdrawWidget.tsx
@@ -92,8 +92,9 @@ export const WithdrawWidget = (props: WithdrawWidgetProps) => {
                         storageTokenDeltas:
                           nep141Storage?.quote?.tokenDeltas ?? [],
                         withdrawParams: (() => {
-                          switch (tokenOut.chainName) {
-                            case "near":
+                          const bridge = tokenOut.bridge
+                          switch (bridge) {
+                            case "direct":
                               return {
                                 type: "to_near",
                                 amount: adjustDecimals(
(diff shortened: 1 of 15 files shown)


src/types/base.ts
```diff
@@ -26,6 +26,8 @@ export type SupportedChainName =
   | "gnosis"
   | "berachain"
 
+export type SupportedBridge = "direct" | "poa" | "aurora_engine"
+
 export interface FungibleTokenInfo extends Partial<BaseTokenBalance> {
```
jobotics, line 31: Is it safe to make all properties of `BaseTokenBalance` optional?

cawabunga-bytes: Not sure if it is even used.

src/features/withdraw/components/WithdrawWidget.tsx
jobotics, line 1: As a suggestion, maybe we can take the `assembleSignMessages` logic and move it to a utility function, something like assembleWithdrawMessages, or move it underneath.

cawabunga-bytes: Yeah I was also thinking about it. Maybe in the future.

src/services/poaBridgeService.ts
```diff
@@ -0,0 +1,52 @@
...
+  )
+
+  return new Promise<WaitForWithdrawalCompletionResult>((resolve, reject) => {
+    const unsubscribe = queryObserver.subscribe((result) => {
+      if (result.data == null) {
+        return
+      }
+
+      // We expect it having a single withdrawal, so we take first
+      const withdrawal = result.data.withdrawals[0]
```
jobotics, line 30: Can there be an empty array here?

cawabunga-bytes: No, that's why we assert it below

jobotics (approved):

PR #367, opened 2025-03-27
fix(account): move label out of link

cawabunga-bytes:


src/features/account/components/SummaryIsland.tsx
```diff
@@ -73,14 +73,14 @@ export function SummaryIsland({
...
           />
-          <NavButton
-            routeName="gift"
-            renderHostAppLink={renderHostAppLink}
-            className="flex-1"
-            variant="secondary"
-            label="Gift"
-            icon={<Gift weight="bold" className="size-5" />}
-          />
+          {/*<NavButton*/}
```
jobotics, line 76: Why do we leave this?

cawabunga-bytes: Because when you finish gifts, then you just uncomment it.

jobotics: Gotcha!

jobotics (approved):

PR #369, opened 2025-03-27
feat: support different timings for quotes

cawabunga-bytes:
Now we request a quote each 3 sec with default solver `wait_ms` value (which is 1.5s). Need to send two parallel requests with 2s and 10s wait time, so MPC solver can be in time.

### DIFF
--- a/src/features/machines/backgroundQuoterMachine.ts
+++ b/src/features/machines/backgroundQuoterMachine.ts
@@ -1,5 +1,4 @@
 import { type ActorRef, type Snapshot, fromCallback } from "xstate"
-import { settings } from "../../constants/settings"
 import { logger } from "../../logger"
 import { type QuoteResult, queryQuote } from "../../services/quoteService"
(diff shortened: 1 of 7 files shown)


src/features/machines/backgroundQuoterMachine.ts
```diff
@@ -47,9 +46,26 @@ type ParentActor = ActorRef<Snapshot<unknown>, ParentEvents>
 
 type Input = {
   parentRef: ParentActor
-  delayMs: number
 }
 
+const QUOTE_TIMINGS = [
```
jobotics, line 51: Maybe it's time to move this to config sdk, I think it's an important part that shouldn't be hidden inside machine, wdyt?

cawabunga-bytes: I don't think that should be configurable. Because it's a part of implementation. I can't find any reasons why anybody wants to pass own configuration.

jobotics (approved):

PR #383, opened 2025-04-01
fix: fix "no quote" for fast quotes

cawabunga-bytes:
PR simplifies polling and fixes various issues introduced with slow quotes:
- race condition between slow and fast quotes
- fast quote often returned "no quote", because of solvers

### DIFF
--- a/src/features/machines/backgroundQuoterMachine.ts
+++ b/src/features/machines/backgroundQuoterMachine.ts
@@ -1,7 +1,13 @@
 import { type ActorRef, type Snapshot, fromCallback } from "xstate"
 import { logger } from "../../logger"
-import { type QuoteResult, queryQuote } from "../../services/quoteService"
+import {
+  type AggregatedQuoteParams,
+  type QuoteResult,
+  queryQuote,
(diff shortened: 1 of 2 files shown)


src/features/machines/backgroundQuoterMachine.ts
```diff
@@ -107,96 +95,102 @@ export const backgroundQuoterMachine = fromCallback<
...
+    onError
+  )
+
+  queryQuote({ ...quoteParams, waitMs: SLOW_QUOTE_WAIT_MS }, { signal }).then(
+    onResult,
+    onError
+  )
+
+  const timer = setInterval(() => {
+    queryQuote({ ...quoteParams, waitMs: SLOW_QUOTE_WAIT_MS }, { signal }).then(
```
jobotics, line 178: Why don't we try to get different quotas in the interval - FAST_QUOTE_WAIT_MS, NORMAL_QUOTE_WAIT_MS , etc.? As I understand right it here we might get faster result, but we wait for MPC solvers for some reason.

jobotics (approved):

PR #389, opened 2025-04-03
refactor: refactor decoding Near RPC query results

cawabunga-bytes:


src/services/blockchainBalanceService.ts
```diff
@@ -18,13 +15,15 @@ export const getNearNativeBalance = async ({
...
   try {
-    const response = await getNearBalance({
+    const response = await nearClient.query({
       request_type: "view_account",
       finality: "final",
       account_id: accountId,
     })
 
-    const balance = BigInt(response.amount)
+    const parsed = v.parse(v.object({ amount: v.string() }), response)
```
jobotics, line 24: Why we can't use `decodeQueryResult` util here?

cawabunga-bytes: Not really, because `query` and `view_account` requests has different responseы. `view_account` returns already decoded data, we just need to transform to bigint. `query` returns raw bytes, which we need to convert to json string and then parse it.

jobotics: I see.

jobotics (approved):

PR #393, opened 2025-04-07
 fix: show "Sign in" button to all widgets for un-authenticated user

jobotics:


src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -520,13 +522,23 @@ export const WithdrawForm = ({
...
+          {isLoggedIn ? (
+            <ButtonCustom
+              size="lg"
+              disabled={state.matches("submitting") || noLiquidity}
+              isLoading={state.matches("submitting")}
+            >
+              {renderWithdrawButtonText(noLiquidity, insufficientTokenInAmount)}
+            </ButtonCustom>
+          ) : (
+            renderHostAppLink(
```
dr-block: any way to avoid this duplication ? 
may be any HOC wrapper ? or anything else ?

jobotics: I didn't get you, could you explain a bit? As `renderHostAppLink` comes form application, in this case this is a Link tag but it could be anything else

dr-block: like we can have CheckHostAppLinkHOC wrapper

```
<CheckHostAppLinkHOC shouldRender={isLoggedIn} renderHostAppLink = renderHostAppLink >
      <ButtonCustom
              size="lg"
              disabled={state.matches("submitting") || noLiquidity}
              isLoading={state.matches("submitting")}
            >
              {renderWithdrawButtonText(noLiquidity, insufficientTokenInAmount)}
       </ButtonCustom>
<CheckHostAppLinkHOC>
```

within HOC component you would decide either render the child or call renderHostAppLink depending on shouldRender condition

jobotics: It isn't a HOC per se, but rather than wrapper component, but I like it

dr-block (approved):

cawabunga-bytes (approved):

PR #394, opened 2025-04-07
refactor: extract intents contract methods to intentsContractService.ts

cawabunga-bytes:


src/services/intentsContractService.ts
jobotics, line 1: We have `intentService.ts` can we or should we move all contract specific functionality from there to here?

cawabunga-bytes: No, this file is a simple wrapper of near client to easy access to contract methods. You can look at it as viem's readContract + abi, we wouldn't need it if near client had better dev experience. It doesn't check anything and barely process data. On other hand intentService will include domain logic and might call intentsContractService, but not the way around.

jobotics (approved):

PR #398, opened 2025-04-08
feat: add extra log for no liquidity when user has enough funds

cawabunga-bytes:
PR adds logging of a new event when user has enough funds, but solvers can't give them quotes.

### DIFF
--- a/src/features/otcDesk/hooks/useOtcTakerConfirmTrade.ts
+++ b/src/features/otcDesk/hooks/useOtcTakerConfirmTrade.ts
@@ -81,7 +81,9 @@ export function useOtcTakerConfirmTrade({
 
       // todo: UI performance: add re-quoting in the background
       // It's totally alright do async stuff after singing
-      const quoteHashesResult = await getFreshQuoteHashes(quotes, quoteParams)
+      const quoteHashesResult = await getFreshQuoteHashes(quotes, quoteParams, {
+        logBalanceSufficient: true,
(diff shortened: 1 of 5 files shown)


src/features/otcDesk/hooks/useOtcTakerConfirmTrade.ts
```diff
@@ -81,7 +81,9 @@ export function useOtcTakerConfirmTrade({
 
       // todo: UI performance: add re-quoting in the background
       // It's totally alright do async stuff after singing
-      const quoteHashesResult = await getFreshQuoteHashes(quotes, quoteParams)
+      const quoteHashesResult = await getFreshQuoteHashes(quotes, quoteParams, {
+        logBalanceSufficient: true,
```
jobotics, line 85: How do we know that the requested swap amount is available in the user's account?

jobotics (approved):

src/features/otcDesk/hooks/useOtcTakerConfirmTrade.ts
cawabunga-bytes: In this case if user reached that far then they have funds (in OTC we quote only if user have funds). I admit this is not very clear, but I didn't come with better solution.

jobotics: I see

PR #404, opened 2025-04-08
refactor: improve parsing RPC response and throwing RPC error

cawabunga-bytes:
Example of errored RPC request:

<p>

Screenshot 2025-04-08 at 18 09 28

</p>
</details>

### DIFF
--- a/src/services/solverRelayHttpClient/apis.ts
+++ b/src/services/solverRelayHttpClient/apis.ts
@@ -5,42 +5,50 @@ export async function quote(
   params: types.QuoteRequest["params"][0],
   config: types.RequestConfig = {}
 ): Promise<types.QuoteResponse["result"]> {
-  const json = await jsonRPCRequest<types.QuoteRequest>("quote", params, config)
-  return json.result
+  const result = await jsonRPCRequest<types.QuoteRequest>(
+    "quote",
+    params,
+    config
+  )
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  return result as any
 }
 
 export async function publishIntent(
   params: types.PublishIntentRequest["params"][0],
   config: types.RequestConfig = {}
 ): Promise<types.PublishIntentResponse["result"]> {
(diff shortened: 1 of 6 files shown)


src/services/solverRelayHttpClient/apis.ts
```diff
@@ -5,42 +5,50 @@ export async function quote(
...
 ): Promise<types.QuoteResponse["result"]> {
-  const json = await jsonRPCRequest<types.QuoteRequest>("quote", params, config)
-  return json.result
+  const result = await jsonRPCRequest<types.QuoteRequest>(
+    "quote",
+    params,
+    config
+  )
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  return result as any
```
jobotics, line 14: Can we improve return type in all such cases?

cawabunga-bytes: We can parse it using valibot. Maybe in the future, but not sure.

src/services/solverRelayHttpClient/runtime.ts
```diff
@@ -1,27 +1,61 @@
+import * as v from "valibot"
 import { config as globalConfig } from "../../config"
+import { handleRPCResponse } from "../../utils/handleRPCResponse"
 import { request } from "../../utils/request"
 import type * as types from "./types"
 
+const rpcResponseSchema = v.union([
```
jobotics, line 7: I'm not sure about the benefits of placing everything in a single schema. But if that wasn't a deliberate decision, it might be better to split it like this:
```
const successResponseSchema = v.object({
  // success schema
})

const errorResponseSchema = v.object({
  // error schema
  ),
})

const rpcResponseSchema = v.union([successResponseSchema, errorResponseSchema])
```

jobotics (approved):

src/services/solverRelayHttpClient/runtime.ts
cawabunga-bytes: It was. Initially I wrote two schemas, but I didn't find it very useful.

PR #405, opened 2025-04-09
fix: withdraw - display bridge's hot wallet's liquidity - ticket: 86erv674w

dr-block:


src/components/Select/Select.tsx
```diff
@@ -12,6 +14,11 @@ type Props<T extends string> = {
       label: string
       icon: ReactNode
       value: string
+      hotBalance?:
```
jobotics: Let's use HotBalanceProps interface here and in places where you provide hotBalance but name it something like `TokenValueWithPrice`

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        if (address != null) {
+          if (balances[address]?.vaultBalance != null) {
+            let price = 1
+            if (tokensUsdPriceData) {
+              price = tokensUsdPriceData[address]?.price ?? 1
+            }
+
+            hotBalance = {
+              amount: BigInt((balances[address] as TokenBalances).vaultBalance),
```
jobotics: Could you check if this type cast is necessary? As balances[address] is already TokenBalances

src/components/Select/HotBalance.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface HotBalanceProps {
+  hotBalance?: TokenValueWithPrice | null
+}
+
+export const HotBalance: FC<HotBalanceProps> = ({ hotBalance }) => {
+  const hotBalanceUSD_1k = hotBalance ? adjustTo1kUsd(hotBalance) : null
+
+  return (
+    <Text className="text-gray-11 text-xs font-medium">
```
cawabunga-bytes: We don't use Text anymore, you can use just div.

src/queries/poaBridgeQueries.ts
```diff
@@ -26,3 +34,19 @@ export function createWithdrawalStatusQueryOptions({
     },
   })
 }
+
+export function useTokenBalancesQueryOptions(
+  addresses: string[],
+  enabled = true
+): UseQueryResult<types.TokenBalancesRequestOk, Error> {
+  const sortedAddressesKey =
+    addresses.length > 0 ? [...addresses].sort().join(",") : ""
```
cawabunga-bytes: What's the purpose of this manipulation?

src/components/Select/Select.tsx
```diff
@@ -12,6 +14,7 @@ type Props<T extends string> = {
       label: string
       icon: ReactNode
       value: string
+      hotBalance?: TokenValueWithPrice | null
```
cawabunga-bytes: That's a generic select, it should not contain domain logic.

jobotics (approved):

src/services/poaBridgeHttpClient/types.ts
```diff
@@ -97,3 +97,22 @@ export type WithdrawalStatusResponseOk = JSONRPCResponse<{
...
+  decimals: number
+  totalSupply: string
+  vaultBalance: string
+  coldWalletBalance: string
+  balanceOnDefuse: string
+}
+
+export type TokenBalancesRequestOk = TokenBalances[]
+
+export type WithdrawalStatusResponseErr = {
```
cawabunga-bytes: That's not needed.

src/services/poaBridgeHttpClient/types.ts
```diff
@@ -97,3 +97,22 @@ export type WithdrawalStatusResponseOk = JSONRPCResponse<{
...
+  address: string
+  nearAddress: string
+  decimals: number
+  totalSupply: string
+  vaultBalance: string
+  coldWalletBalance: string
+  balanceOnDefuse: string
+}
+
+export type TokenBalancesRequestOk = TokenBalances[]
```
cawabunga-bytes: Let's name it just BridgeBalanceResponse or something. Because it's not "request" and "tokenBalance" is too generic (like is it user balance or what)

src/services/poaBridgeHttpClient/runtime.ts
```diff
@@ -56,3 +56,13 @@ export async function jsonRPCRequest<
...
+
+export async function jsonConsoleBalancesRequest(params: string) {
+  const response = await request({
+    url: `${globalConfig.env.poaBridgeBaseURL}/console/tokenBalances?${params}`,
+    fetchOptions: {
+      method: "GET",
+    },
+  })
+  return response.json()
+}
```
cawabunga-bytes: You don't need this function, it won't be reused so you can just inline it. The same as src/services/tokensUsdPricesHttpClient/apis.ts

src/services/poaBridgeHttpClient/apis.ts
```diff
@@ -52,3 +52,16 @@ export async function getWithdrawalStatus(
...
 }
+
+export async function getTokenBalancesRequest(
+  addresses: string[]
+): Promise<types.TokenBalancesRequestOk> {
+  const params = addresses.map((address) => `addresses[]=${address}`).join("&")
+  const balances = await jsonConsoleBalancesRequest(params)
+
+  if ("error" in balances) {
+    throw new Error(balances.error)
```
cawabunga-bytes: When this can appear?

src/services/poaBridgeHttpClient/apis.ts
```diff
@@ -52,3 +52,16 @@ export async function getWithdrawalStatus(
...
+  }
+
+  return balances || []
```
cawabunga-bytes: When there might be no value?

src/types/base.ts
```diff
@@ -59,3 +59,7 @@ export interface TokenValue {
   amount: bigint
   decimals: number
 }
+
+export interface TokenValueWithPrice extends TokenValue {
```
cawabunga-bytes: I doubt it is a base type.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const onlyPoaToken = filterOutPoaBridgeTokens(token)
+  const addresses = onlyPoaToken
+    ? getTokenAssetIdsWithoutNep141(onlyPoaToken)
+    : []
+  const { amount: userBalance, decimals: userBalanceDecimals } =
+    tokenInBalance || { amount: 0n, decimals: 1 }
+  let showHotBalances = userBalance > 0
+
+  const { data } = useTokenBalancesQueryOptions(addresses, showHotBalances)
```
cawabunga-bytes: This part is worth to move to own function (or to `useTokenBalancesQueryOptions`):
```ts
  const onlyPoaToken = filterOutPoaBridgeTokens(token)
  const addresses = onlyPoaToken
    ? getTokenAssetIdsWithoutNep141(onlyPoaToken)
    : []
  const { amount: userBalance, decimals: userBalanceDecimals } =
    tokenInBalance || { amount: 0n, decimals: 1 }
  let showHotBalances = userBalance > 0

  const { data } = useTokenBalancesQueryOptions(addresses, showHotBalances)
```

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+        acc[curr.chainName] = curr.defuseAssetId
+        return acc
+      }, {})
+
+  const onlyPoaToken = filterOutPoaBridgeTokens(token)
+  const addresses = onlyPoaToken
+    ? getTokenAssetIdsWithoutNep141(onlyPoaToken)
+    : []
+  const { amount: userBalance, decimals: userBalanceDecimals } =
+    tokenInBalance || { amount: 0n, decimals: 1 }
```
cawabunga-bytes: Please avoid "||" as sometimes it might works unexpectedly (e.g. empty string). Use "??" instead.

src/utils/tokenUtils.ts
```diff
@@ -352,3 +358,58 @@ export function accountSlippageExactIn(
     return [token, amount]
   })
 }
+
+export function filterOutPoaBridgeTokens(
```
cawabunga-bytes, line 356: I think it should more generic function with signature:
```ts
filterTokens(t: BaseTokenInfo | UnifiedTokenInfo, (t: BaseTokenInfo) => bool): BaseTokenInfo[]
```
Then you usage would be:
```ts
filterTokens(token, (t) => t.bridge === 'poa')
```

src/queries/poaBridgeQueries.ts
```diff
@@ -26,3 +34,19 @@ export function createWithdrawalStatusQueryOptions({
     },
   })
 }
+
+export function useTokenBalancesQueryOptions(
```
cawabunga-bytes: nit: Function name is confusing. Also it's not query options.

src/queries/poaBridgeQueries.ts
```diff
@@ -26,3 +34,19 @@ export function createWithdrawalStatusQueryOptions({
...
+  addresses: string[],
+  enabled = true
+): UseQueryResult<types.TokenBalancesRequestOk, Error> {
```
cawabunga-bytes: nit: Return type is unnecessary.

src/utils/tokenUtils.ts
```diff
@@ -352,3 +358,58 @@ export function accountSlippageExactIn(
...
+    return token.bridge === "poa" ? token : null
+  }
+
+  return {
+    ...token,
+    groupedTokens: token.groupedTokens.filter((t) => t.bridge === "poa"),
+  }
+}
+
+export function getTokenAssetIdsWithoutNep141(
```
cawabunga-bytes: nit: it's `getTokenAccountIds`

src/utils/tokenUtils.ts
```diff
@@ -352,3 +358,58 @@ export function accountSlippageExactIn(
...
+  }
+
+  return token.groupedTokens.map((t) => {
+    return t.defuseAssetId.startsWith(stringCleaner)
+      ? t.defuseAssetId.replace(stringCleaner, "")
+      : t.defuseAssetId
+  })
+}
+
+export function addNep141ToAddress(address: string): string {
```
cawabunga-bytes: nit: To be more generic it's `tokenAccountIdToDefuseAssetId` or just `toDefuseAssetId`

src/utils/tokenUtils.ts
```diff
@@ -352,3 +358,58 @@ export function accountSlippageExactIn(
...
+  const nep141 = "nep141:"
+
+  if (!address || address.toLowerCase().startsWith(nep141)) {
```
cawabunga-bytes: I don't think anybody should call this function on defuse asset id. It either should throw or do not check it at all.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+  const balances =
+    data?.reduce((acc: { [key: string]: TokenBalances }, curr) => {
+      if (
+        compareAmounts(
+          { amount: userBalance, decimals: userBalanceDecimals },
+          { amount: BigInt(curr.vaultBalance), decimals: curr.decimals }
+        ) > 0
+      ) {
+        // we should show in case user's balance is MORE than any of the HOT chain balances
+        anyHotBalanceIsLessThanUserBalance = true
```
cawabunga-bytes: nit: That's weird use of `reduce`, since you have a side-effect inside. Personally I don't like using `reduce` at all, I prefer for-of or map.

cawabunga-bytes: Another option would be 

```ts
data.some((t) => compare(tokenInBalance, {amount: BigInt(t.vaultBalance), decimals: t.decimals}))
```

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+        return acc
+      }, {})
+
+  const onlyPoaToken = filterOutPoaBridgeTokens(token)
+  const addresses = onlyPoaToken
+    ? getTokenAssetIdsWithoutNep141(onlyPoaToken)
+    : []
+  const { amount: userBalance, decimals: userBalanceDecimals } =
+    tokenInBalance || { amount: 0n, decimals: 1 }
+  let showHotBalances = userBalance > 0
```
cawabunga-bytes: You're mixing domain and UI logic here. I would suggest removing `show*` variables, and replace them with better declarative one, for example, userHasMoreFundsThanBridgeHas (of course not this one, but it represents the idea). And then components decide weather they want to render or no.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
cawabunga-bytes: You are having unnecessary waterfall requests. You wait until users balance is loaded.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+  showHotBalances = showHotBalances && anyHotBalanceIsLessThanUserBalance
+
+  const blockchainSelectItems = Object.fromEntries(
+    allBlockchains
+      .filter((blockchain) => availableBlockchains[blockchain.value])
+      .map((a) => {
+        const address = availableBlockchains[a.value]
+        let hotBalance = null
+
+        if (address != null) {
```
cawabunga-bytes: Also it should be always defined. It's better to just throw an error if it doesn't.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+          if (balances[address]?.vaultBalance != null) {
+            let price = 1
+            if (tokensUsdPriceData) {
+              price = tokensUsdPriceData[address]?.price ?? 1
```
cawabunga-bytes: Why price is 1? What are consequences of that?

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            const addressBalance = balances[address]
+            assert(addressBalance != null)
```
cawabunga-bytes: You can avoid this. If you store bridge token info in a variable:
```ts
const balance = balances[address]
if (balance != null) {}

// instead of existing
if (balances[address]?.vaultBalance != null) {
```

cawabunga-bytes (changes requested):

src/queries/poaBridgeQueries.ts
dr-block: - sorting in case we have same addresses but different order 
- [...] for not manipulating addresses instance as we run sort on it
- joining as a string to not have a lot of keys but have one key (optional)

src/utils/tokenUtils.ts
dr-block: it can be, but prefer having it simple and on purpose, let it do one thing and do it .
in case other filters, we can have other logic and separate functions ...

src/utils/tokenUtils.ts
dr-block: we loose nothing I guess keeping this additional check ....

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
dr-block: I also think it is doing 2 things, let's go with for of

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
dr-block: so I thought if no price ... what should do in that case... thought 1 is secure , so let's say if that is turbo and we do not have price for turbo to convert to $ , what to do ?

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
dr-block: think the new approach suits, lmk it is or not ...

src/components/Select/HotBalance.tsx
cawabunga-bytes, line 1: This is also should be moved somewhere else, maybe to the withdraw feature.

src/features/withdraw/components/WithdrawForm/LongWithdrawWarning/index.tsx
```diff
@@ -2,28 +2,51 @@ import { Callout } from "@radix-ui/themes"
...
+  const shouldShow =
+    Number(formatUnits(amountIn.amount, amountIn.decimals)) * tokenPrice >=
+    cleanerHotBalance
+
+  if (!shouldShow) return null
+
+  const adjustedBalance = adjustTo1kUsd(hotBalance)
+  const adjustedBalanceText =
+    adjustedBalance === 0
+      ? `Unfortunately ${adjustedBalance}`
```
cawabunga-bytes, line 39: Did you miss "k"?

src/utils/tokenUtils.ts
cawabunga-bytes: We don't manipulate the token object in that way you did it. For me it looks fragile. I suggest returning just an array of BaseTokenInfo's.

src/utils/tokenUtils.ts
cawabunga-bytes: You can pass incorrect data and then it won't yell at you at all. It isn't robust approach. Let's do it in correct way.

src/features/withdraw/components/WithdrawForm/LongWithdrawWarning/index.tsx
dr-block: nope in the case when the hot balance is 0 I adjusted the message a bit, that we would say Unfortunetly 0 is available as instant checkout .... so I think no need to have 0k

src/queries/poaBridgeQueries.ts
cawabunga-bytes: `addresses.slice().sort()` will be more perfomant (overall it's better to to use `arr.toSorted()` but it has 94% support rate). And also you don't need manually convert to string, react query will do it for you.

cawabunga-bytes: Just for you information, objects in the key are hashed deterministically (not arrays though)

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
cawabunga-bytes: That was the question, why we need 1 and what will happen if there will be null/undefined?

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,41 @@
...
+  const availableBlockchains = getAvailableBlockchains(token)
+
+  return Object.fromEntries(
+    allBlockchains
+      .filter((blockchain) => availableBlockchains[blockchain.value])
+      .map((a) => {
+        const address = availableBlockchains[a.value]
+        assert(address != null)
+
+        let hotBalance = null
```
cawabunga-bytes: Not sure why ts didn't say anything. Type is missed here.

src/features/withdraw/components/WithdrawForm/hooks/useBlockchainSelectItems.ts
```diff
@@ -0,0 +1,41 @@
+import type { TokenUsdPriceData } from "../../../../../hooks/useTokensUsdPrices"
+import type { TokenBalances } from "../../../../../services/poaBridgeHttpClient/types"
+import type { SwappableToken } from "../../../../../types/swap"
+import { assert } from "../../../../../utils/assert"
+import { allBlockchains } from "../constants"
+import { getAvailableBlockchains } from "../utils"
+
+export const useBlockchainSelectItems = (
```
cawabunga-bytes: That's not a hook, it's just a regular function.

src/features/withdraw/components/WithdrawForm/hooks/useTokenBalances.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { SwappableToken } from "../../../../../types/swap"
+import { tokenAccountIdToDefuseAssetId } from "../../../../../utils/tokenUtils"
+
+export const useTokenBalances = (
+  token: SwappableToken,
+  hasAnyBalance: boolean
+) => {
+  const { data } = useTokenBalancesQuery(token, hasAnyBalance)
+
+  const balances: { [key: string]: TokenBalances } = {}
```
cawabunga-bytes: nit: `Record<string, TokenBalances>`

src/features/withdraw/components/WithdrawForm/hooks/useTokenBalances.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (data) {
+    for (const balance of data) {
+      balances[tokenAccountIdToDefuseAssetId(balance.nearAddress)] = balance
+    }
```
cawabunga-bytes, line 16: nit: Since useTokenBalancesQuery is used in one place (and probably won't be used anywhere else), then it probably better to move data processing to "select" in useQuery, so you won't have this hook at all.

cawabunga-bytes (approved):
Good work! Please address rest comments and good to go!

PR #414, opened 2025-04-10
feat: improve quotes loading status

cawabunga-bytes:
1. We don't want to show error if 2s quote fails
2. We want show extra message when quote takes long time

<p>

Screenshot 2025-04-10 at 17 20 25

</p>
</details>

### DIFF
--- a/src/components/Form/FieldComboInput.tsx
+++ b/src/components/Form/FieldComboInput.tsx
@@ -1,6 +1,6 @@
 import { Skeleton } from "@radix-ui/themes"
 import clsx from "clsx"
-import { useRef } from "react"
+import { useEffect, useRef, useState } from "react"
 import type {
   FieldError,
   FieldErrors,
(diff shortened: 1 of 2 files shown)


src/components/Form/FieldComboInput.tsx
```diff
@@ -170,3 +183,22 @@ export const FieldComboInput = <T extends FieldValues>({
 }
 
 FieldComboInput.displayName = FieldComboInputRegistryName
+
+/**
+ * Sets a value after a delay
+ */
+function useThrottledValue<T>(value: T, delayMs: number): T {
```
jobotics, line 190: I think this should be separate hook as this might be useful not only with `FieldComboInput`

cawabunga-bytes: If somebody wants to use it, then they will need to extract it.

jobotics: It's possible, but this hook isn't really tied to this component

jobotics (approved):

src/components/Form/FieldComboInput.tsx
cawabunga-bytes: Yeah that's true, but I still prefer collocation when possible.

PR #435, opened 2025-04-18
fix: add hot balance support for non poa bridge chains ticket: 86et6g61g

dr-block:


src/features/withdraw/components/WithdrawForm/constants.tsx
```diff
@@ -148,3 +149,5 @@ export const allBlockchains = [
   icon: ReactNode
   value: SupportedChainName
 }>
+
+export const SolverId = "solver-multichain-asset.near" as IntentsUserId // currently we have only this multychain solver
```
cawabunga-bytes, line 153: That will work in short term, but won't in the long run. Last week I found out that **anyone** can become a solver without any requirements. So your approach won't work when we don't even know who solvers are.

cawabunga-bytes (approved):
I am approving current approach, but it very short term solution.

PR #442, opened 2025-04-23
feat: show PoA withdrawal fee ticket: 86ertnjha

dr-block:


src/services/withdrawService.ts
```diff
@@ -75,15 +80,46 @@ export async function prepareWithdraw(
     depositedBalanceRef,
     poaBridgeInfoRef,
     backgroundQuoteRef,
+    userAddress,
   }: {
     formValues: WithdrawFormContext
     depositedBalanceRef: ActorRefFrom<typeof depositedBalanceMachine>
     poaBridgeInfoRef: ActorRefFrom<typeof poaBridgeInfoActor>
     backgroundQuoteRef: ActorRefFrom<typeof backgroundQuoterMachine>
+    userAddress: string | null
```
cawabunga-bytes, line 101: Unauthenticated user should not rich this point, do they?

src/services/withdrawService.ts
```diff
@@ -75,15 +80,46 @@ export async function prepareWithdraw(
...
+    return {
+      tag: "err",
+      value: { reason: "ERR_BALANCE_FETCH" },
+    }
+  }
+
+  const balances = balances_.value
+  const withdrawalEstimateData =
+    withdrawalEstimateData_.status === "rejected"
+      ? null
```
cawabunga-bytes: You decided not to handle failed promise?

src/services/withdrawService.ts
```diff
@@ -186,6 +223,12 @@ export async function prepareWithdraw(
       swap: swapRequirement,
       nep141Storage: nep141Storage.value,
       receivedAmount: receivedAmount,
+      withdrawalEstimateData: withdrawalEstimateData
```
cawabunga-bytes: nit: `withdrawalEstimateData` is not clear name, just withdtrawalFee would be fine?

src/services/withdrawService.ts
```diff
@@ -186,6 +223,12 @@ export async function prepareWithdraw(
...
+        ? {
+            amount: BigInt(withdrawalEstimateData.withdrawalFee),
+            decimals: withdrawalEstimateData.withdrawalFeeDecimals,
+          }
+        : null,
```
cawabunga-bytes: Instead of null you can return 0 value.

src/services/withdrawService.ts
dr-block: Unauthenticated user is reaching currently to this state, more than that, unauthenticated user can type any value , choose any token, choose any chain
BUT, calculation would end up in the infinite loading ...

src/services/withdrawService.ts
dr-block: Failed promise was equalized to the cases when we do not have any information about fee. 
cause actually we have 3 cases 
- fee  - showing { fee Token }
- fee is 0 - showing {0 Token}
- we have no information ... (I didn't mark that as 0 for now, I thought it is OK to show { - Token})

src/services/withdrawService.ts
dr-block: agree

src/services/withdrawService.ts
dr-block: the interesting case is that I decided to have 3. cases 
- fee - showing { fee Token }
- fee is 0 - showing {0 Token}
- we have no information ... (I didn't mark that as 0 for now, I thought it is OK to show { - Token})

src/services/withdrawService.ts
cawabunga-bytes: I think it is better to have a special error case when we can't get withdrawal fee (also log it).

src/services/withdrawService.ts
cawabunga-bytes: If you return early with error when you can't compute withdrawal fee, then at this point you will always have data.

cawabunga (dismissed):

cawabunga-bytes (approved):

PR #445, opened 2025-05-05
feat: add presets from query params to the withdraw widget

FroggyRocky:
Accept query parameters for withdraw preset configs on redirect from Aurora Cloud Console gas abstraction transactions top-up

### DIFF
--- a/src/features/withdraw/components/WithdrawForm/index.tsx
+++ b/src/features/withdraw/components/WithdrawForm/index.tsx
@@ -39,7 +39,10 @@ import type {
   TokenValue,
   UnifiedTokenInfo,
 } from "../../../../types/base"
-import type { WithdrawWidgetProps } from "../../../../types/withdraw"
+import {
+  type WithdrawWidgetProps,
+  isSupportedChainName,
+} from "../../../../types/withdraw"
 import { parseUnits } from "../../../../utils/parse"
 import { validateAddress } from "../../../../utils/validateAddress"
 import {
@@ -89,6 +92,9 @@ export const WithdrawForm = ({
   userAddress,
   chainType,
   tokenList,
+  presetAmount,
+  presetNetwork,
+  presetRecipient,
   sendNearTransaction,
   renderHostAppLink,
 }: WithdrawFormProps) => {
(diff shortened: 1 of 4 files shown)


src/components/Modal/ModalSelectAssets.tsx
```diff
@@ -151,7 +151,7 @@ export const ModalSelectAssets = () => {
           <div className="flex flex-col gap-4">
             <div className="flex flex-row justify-between items-center">
               <Text size="5" weight="bold">
-                Select asset
+                SELECT ASSSETTTSSAAA
```
amdefuse: I assume this is some leftover from testing 😅

src/types/withdraw.ts
```diff
@@ -17,3 +25,27 @@ export type WithdrawWidgetProps = {
...
+    "optima",
+    "coineasy",
+    "aurora",
+    "xrpledger",
+    "zcash",
+    "gnosis",
+    "berachain",
+    "tron",
+  ].includes(chainName)
+}
```
amdefuse: I wonder if we can pull this list from some other place where they're already defined in the app. Might need some guidance from someone that knows the code better but a quick search suggests we might already be maintaining this sort of list in multiple places 🤔 

https://github.com/search?q=repo%3Adefuse-protocol%2Fdefuse-sdk%20coineasy&type=code

src/features/withdraw/components/WithdrawForm/index.tsx
```diff
@@ -327,6 +332,18 @@ export const WithdrawForm = ({
...
+    if (presetAmount != null) {
+      setValue("amountIn", presetAmount)
+    }
+    if (presetNetwork != null && isSupportedChainName(presetNetwork)) {
+      setValue("blockchain", presetNetwork)
+    }
+    if (presetRecipient != null) {
+      setValue("recipient", presetRecipient)
+    }
+  }, [presetAmount, presetNetwork, presetRecipient, setValue])
```
amdefuse, line 341: I wonder if we could set these via the [defaultValues](https://react-hook-form.com/docs/useform#defaultValues) option of `react-hook-form`.

src/components/Modal/ModalSelectAssets.tsx
daniladefuse: you are right, sorry for that!

src/types/withdraw.ts
hyper54n3: I would move this to ´src/utils/blockchain.ts´,which is an attempt to centralize this scattering that @amdefuse mentioned.

hyper54n3 (approved):
Approved with comments

PR #447, opened 2025-05-12
feat: make OTC generate links async

jobotics:
Enable async OTC link generation to store links in the client DB and use trade_id in queries instead of long hashes.


src/features/otcDesk/hooks/useGenerateOrderLink.ts
```diff
@@ -0,0 +1,49 @@
...
+      } finally {
+        if (isMounted) {
+          setIsLoading(false)
+        }
+      }
+    }
+
+    generateOrderLink()
+    return () => {
+      isMounted = false
```
dr-block: can you pls explain why do we need this isMounted ? 
as seems isMounted would be always true in this useffect ...
may be let isMounted = true should be written OUT of the useffect ?

src/features/otcDesk/actors/otcMakerReadyOrderActor.ts
```diff
@@ -30,13 +30,15 @@ export type OTCMakerReadyOrderActorInput = {
   usedNonceBase64: string
   multiPayload: MultiPayload
   signerCredentials: SignerCredentials
+  pKey: string
```
dr-block, line 33: can we have more self explanative name than pKey ?

src/features/otcDesk/actors/otcMakerRootMachine.ts
```diff
@@ -32,46 +34,79 @@ import {
...
+  type: "COMPLETE_STORING"
+  multiPayload: MultiPayload
+  signerCredentials: SignerCredentials
+  usedNonceBase64: string
+  tradeId: string
+  pKey: string
+}
+
+type InputType = {
+  tokenList: (BaseTokenInfo | UnifiedTokenInfo)[]
```
dr-block, line 61: we have SwappableToken , would that work instead BaseTokenInfo | UnifiedTokenInfo ? (here and below)

jobotics: I didn’t refactor or change any of the types—just moved them out of the setup() call to reduce the complexity TypeScript has to infer and serialize. This is necessary to resolve the TS7056 error.

It would work, but I don’t think we need to refactor it in the context of this task cause it will required change it not only in this file.

src/features/otcDesk/actors/otcMakerReadyOrderActor.ts
jobotics: This is common abbreviations in cryptography, same as for `iv` (initialization vector), which we use so I would stay with it.

dr-block (approved):

cawabunga-bytes (approved):
all good!

PR #448, opened 2025-05-13
fix: prevent submission tx on preparation

jobotics:


src/features/deposit/components/DepositForm/index.tsx
```diff
@@ -129,9 +130,19 @@ export const DepositForm = ({
...
+      "editing" in snapshot.value &&
+      snapshot.value.editing !== "preparation"
+
+    if (isFormReadyForSubmission) {
+      depositUIActorRef.send({
+        type: "SUBMIT",
+      })
+    } else {
+      logger.warn("Deposit preparation is not complete")
+    }
```
cawabunga-bytes: This logic should be in the machine itself. (in guard probably)

jobotics: Yes, I tried to do this first, but I couldn't find a way how to get machine state itself in a guard, am I missing something?

jobotics: You are right, we can use `stateIn`

cawabunga-bytes (approved):

src/features/machines/depositUIMachine.ts
```diff
@@ -294,49 +295,64 @@ export const depositUIMachine = setup({
         SUBMIT: [
           {
             target: "submittingNearTx",
-            guard: "isChainNearSelected",
+            guard: and(["isChainNearSelected", stateIn("editing.idle")]),
```
dr-block: why not having a func preventIdle(guards) => {.return [...guards, stateIn("editing.idle")] }

dr-block (approved):

src/features/machines/depositUIMachine.ts
jobotics: used a more reliable way

PR #451, opened 2025-05-14
fix: new gen virtual chains deposits

hyper54n3:
Fixes the issue with Tuxachain, Optima, Vertex and Coineasy.

Implements `exitToNearPrecompile` for native tokens and `auroraErc20ABI` for deposits.


src/utils/blockchain.ts
```diff
@@ -78,3 +78,462 @@ export function isSupportedChainName(
   const supportedChainNames = Object.keys(CHAIN_IDS)
   return supportedChainNames.includes(chainName)
 }
+
+export const auroraErc20ABI = [
```
dr-block, line 82: is not it the same like in artifacts/auroraErc20ABI.json file ? if yes, can we reuse ?

hyper54n3: Ah yes, that was a leftover. Deleted. Thanks!

jobotics:
Good job!

jobotics (approved):

PR #454, opened 2025-05-20
Fix zero liquidity on eth near

dr-block:


src/services/deprecatedTokensService.ts
```diff
@@ -0,0 +1,29 @@
...
+      )
+    }
+
+    return DeprecatedTokensService.#instance
+  }
+
+  public getValidToken = (address: string): string => {
+    return this.deprecatedTokenToValidToken?.[address] ?? address
+  }
+}
```
cawabunga-bytes: This service does not make sense for me from different perspectives. 

1. Why such implementation of the singleton from Java/Php
2. Instantiation (`DeprecatedTokensService.makeInstance(props.deprecatedTokenToReplacementList)`) behaves unexpectedly if you try to create a new instance with different arguments, you end up with old instance and new list will be discarded. This is counter intuitive.
3. Not every deprecated token may have valid version.

src/types/withdraw.ts
```diff
@@ -19,4 +21,5 @@ export type WithdrawWidgetProps = {
    * Prop is not reactive, set it once when the component is created.
    */
   referral?: string
+  deprecatedTokenToReplacementList?: DeprecatedTokenToReplacedToken
```
cawabunga-bytes: Person who is not familiar with the original problem will never understand what this prop does.

cawabunga-bytes (commented):
I got your idea, but I think we need to have a transparent version of it. I suggest implementing straight forward solution. Just eject broken ETH version from the unified token list to own token object with `ETH (old)` symbol or something.

src/services/deprecatedTokensService.ts
dr-block: 1. why not ? service is service, it can use the class based approach , why not ?
2. that is the reason of singleton , your business logic is that you have one instance at an app scope
3. that's a case that should be handled when it appear, that time this service can be extended 

let me know if you still expect something else to see, if yes, pls explain what are you expectations ?

src/services/deprecatedTokensService.ts
```diff
@@ -0,0 +1,82 @@
...
+    return DeprecatedTokensService.#instance
+  }
+
+  /**
+   * Returns the existing singleton instance of `DeprecatedTokensService`.
+   *
+   * @returns The singleton instance of `DeprecatedTokensService`.
+   */
+  public static getInstance(): DeprecatedTokensService {
+    return DeprecatedTokensService.#instance
```
jobotics: This can throw `TypeError` if `makeInstance()` wasn't called first, suggestion:
```
if (!DeprecatedTokensService.#instance) {
    throw new Error("DeprecatedTokensService not initialized. Call makeInstance() first.")
}
```

src/services/deprecatedTokensService.ts
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * @param deprecatedTokenToValidToken - Optional mapping used during first initialization.
+   * @returns The singleton instance of `DeprecatedTokensService`.
+   */
+  public static makeInstance(
+    deprecatedTokenToValidToken?: DeprecatedTokenToReplacedToken
+  ): DeprecatedTokensService {
+    if (!DeprecatedTokensService.#instance) {
+      DeprecatedTokensService.#instance = new DeprecatedTokensService(
+        deprecatedTokenToValidToken ?? null
```
jobotics: Nit: `?? null` isn't needed since the constructor already defaults to `null`.

src/services/deprecatedTokensService.ts
jobotics: It seems like this adds unnecessary complexity for relatively simple logic. A small, stateless function like getTokenReplacement could achieve the same goal with clearer behavior and easier testing.

cawabunga-bytes (approved):

jobotics (approved):

PR #458, opened 2025-05-21
feat: mirror src folder structure of built library

cawabunga-bytes:
PR changes how "dist" folder look like. Previously there were a couple of bundled files. This PR makes dist folder to mirror src folder. In that way it is easier to access internal functions from outside.

<details><summary>dist folder screenshot</summary>
<p>

Screenshot 2025-05-21 at 15 34 30

</p>
</details>


tsconfig.json
```diff
@@ -1,17 +1,18 @@
...
-    "noEmit": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
     "strict": true,
     "esModuleInterop": true,
     "module": "esnext",
     "moduleResolution": "node",
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
-    "declarationMap": true,
```
jobotics, line 14: What is the reason to disable source maps?

jobotics (approved):

tsconfig.json
```diff
@@ -1,17 +1,18 @@
...
   "compilerOptions": {
     "baseUrl": ".",
-    "outDir": "./dist",
+    "rootDir": "src",
+    "outDir": "dist",
     "target": "es2020",
     "declaration": true,
-    "noEmit": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
```
jobotics, line 8: Let's resolve conflicting options `noEmit` and `emitDeclarationOnly`

tsconfig.json
cawabunga-bytes: To be honest I am not even sure why sourcemaps of type exist. 

https://www.totaltypescript.com/tsconfig-cheat-sheet says:
```
    /* AND if you're building for a library in a monorepo: */
    "composite": true,
    "declarationMap": true,
```

And since we previously we didn't have `d.ts.map` and we aren't building a monorepo lib then I decided to ditch it.

PR #463, opened 2025-05-27
feat: fast withdrawal - new approach

dr-block:


src/types/withdraw.ts
```diff
@@ -4,6 +4,24 @@ import type { BaseTokenInfo, UnifiedTokenInfo } from "./base"
...
+  amount: {
+    value: string
+    __type?: string
+  }
+  lastStepSize?: {
+    value: string
+    __type?: string
+  }
+  lastLiquidityCheck?: LastLiquidityCheckStatus
+}
```
cawabunga-bytes: I think such details should be handled on the higher level. At this point data should be already formatted in consumable structure.

src/features/withdraw/components/WithdrawForm/utils.ts
```diff
@@ -257,3 +210,160 @@ export const getWithdrawButtonText = (
...
+      const { defuseAssetId: defuseAssetIdFrom, decimals: decimalsFrom } =
+        tokenFrom
+
+      const onUserAmount = {
+        amount: addressToUserBalance[defuseAssetIdFrom]?.amount ?? 0n,
+        decimals: decimalsFrom,
+      }
+
+      const liquidity =
+        liquidityData?.[`${defuseAssetIdFrom}#${defuseAssetIdTo}`]
```
cawabunga-bytes, line 324: Maybe it is better to have nested object instead of weird key?
```ts
liquidityData[token1][token2]=data
```

cawabunga-bytes (commented):
Great work! I left some comments.

jobotics (approved):

jobotics:
LGTM

src/features/withdraw/components/WithdrawForm/utils.ts
dr-block: It is actually the way I keep data in "{from token}#{to token}" combination ... 

May be your suggestion is better, but currently nesting would mean to restructure all existing data in the Redis and start supporting new approach .... so let's for now keep the way it works unless we see a huge benefit to use nesting ..

thanks

cawabunga-bytes (approved):

src/sdk/poaBridge/intentHttpClient/api.ts
cawabunga-bytes, line 1: But it is not part of poa bridge.

src/queries/solverLiquidityQuerires.ts
```diff
@@ -0,0 +1,9 @@
+import { useQuery } from "@tanstack/react-query"
+import { getSolverLiquidityRequest } from "../sdk/poaBridge/intentHttpClient/api"
+
+export function useSolverLiquidityQuery() {
+  return useQuery({
+    queryKey: ["solver_liquidity"],
+    queryFn: getSolverLiquidityRequest,
+  })
+}
```
cawabunga-bytes, line 44: nit: `Querires` -> `Queries`

Also I would suggest writing useQuery right in the place where it is used, since it is not complex function and it is not reused anywhere else.

src/features/withdraw/components/WithdrawForm/hooks/useLiquidityData.ts
```diff
@@ -0,0 +1,26 @@
...
+      const liquidityData_: Record<string, bigint> = {}
+
+      for (const { address_from, address_to, validated_amount } of data) {
+        liquidityData_[`${address_from}#${address_to}`] =
+          BigInt(validated_amount)
+      }
+
+      setLiquidityData(liquidityData_)
+    }
+  }, [data, isLoading])
```
cawabunga-bytes: You can move this to query "select:" options.

cawabunga-bytes (changes requested):

src/config.ts
```diff
@@ -20,13 +21,16 @@ const configsByEnvironment: Record<NearIntentsEnv, EnvConfig> = {
     poaBridgeBaseURL: "https://bridge.chaindefuser.com",
     solverRelayBaseURL: "https://solver-relay-v2.chaindefuser.com",
     managerConsoleBaseURL: "https://api-mng-console.chaindefuser.com/api/",
+    // nearIntentsBaseURL: "https://app.near-intents.org/api/",
+    nearIntentsBaseURL: "http://localhost:3000/api/",
```
cawabunga-bytes: Please remove this.

cawabunga-bytes (changes requested):

src/features/withdraw/components/WithdrawForm/hooks/useLiquidityData.ts
cawabunga-bytes, line 1: This hook is useless. I suggest getting rid of it.

cawabunga-bytes (changes requested):

cawabunga-bytes (approved):

PR #466, opened 2025-05-28
feat: add support for HOT bridge; add Polygon and BSC

cawabunga-bytes:
PR introduces a few new features:
- Add Polygon and BSC chains
- Add support for HOT Bridge (deposits and withdraws)

Other important fixes:
- Handling defuse asset ids, as HOT bridge uses nep245 standard instead of nep141. But in our code we assumed that nep141 was the only option.
- Fixes build in watch mode, it is used to be `yarn build -w`, but because build now includes two commands rollup and tsc, we need to run in watch mode both these commands. `yarn build:watch` is the fix.

# HOT Bridge details

As already mentioned above HOT Bridge (aka HOT Omni Bridge) uses NEP-245 standard. In short it is a multi-token standard compared to NEP-141, where one contract = one token. NEP-245 allows having multiple tokens in one contract, which differentiated by internal token ids. Here is examples of hot bridged tokens inside Near Intents:
- `nep245:v2_1.omni.hot.tg:137_11111111111111111111`, where prefix `nep245` is a token standard, `v2_1.omni.hot.tg` is contract account id and `137_11111111111111111111` internal token id (HOT team choose token ids, not we). This asset is native token on Polygon (id = 137).
- `nep245:v2_1.omni.hot.tg:137_qiStmoQJDQPTebaPjgx5VBxZv6L`,  this is USDC on Polygon

<details><summary>Diagram: How withdrawal using HOT Bridge works</summary>
<p>

Untitled

</p>
</details> 

As HOT Bridge does not provide "passive" deposit and "passive" withdrawals, our team created a service that does this. It is wrapped with PoA Bridge. So here is how make a deposit:
1. Generate deposit address using PoA Bridge (absolutely the same as regular PoA deposit)
2. Transfer funds to the address
3. Auto-receive funds (NEP-245 tokens) on Near Intents.

how to make a withdrawal:
1. Send mt_withdraw intent
2. Call `hot_omni_withdraw` to notify backend to process withdraw. <--- this step will be improved in the future
3. Auro-receive funds on destination chain

HOT Bridge has SDK but it is still rough, so only a few functions are used from it.


yarn.lock
jobotics, line 1: Are all the yarn.lock changes due to the new libs we added?

jobotics (approved):

yarn.lock
cawabunga-bytes: Yep, I'm also overwhelmed

PR #467, opened 2025-05-29
feat: include IV in gifts

jobotics:


src/features/gift/types/sharedTypes.ts
```diff
@@ -15,8 +16,22 @@ export type GiftLinkData = {
...
   signerCredentials: SignerCredentials
   signatureResult: WalletSignatureResult
 }
+
+export type CreateGiftIntent = (
+  payload: GiftLinkData
+) => Promise<{ iv: string }>
+
+export type GenerateLink = (params: {
+  secretKey: string
```
cawabunga-bytes, line 29: nit: it's privateKey (or pKey as you already used somewhere)

cawabunga-bytes (approved):

src/features/gift/types/sharedTypes.ts
jobotics: It's a bit confusing, these are distinct cryptographic primitives with different purposes:
- in the otc, pKey represents an AES-GCM symmetric encryption key
- in the gift, secretKey represents an Ed25519 private key in format `Ed25519:privateKey`

So, I think it's better to leave it as is.

jobotics: And, this secretKey and message will be encrypted in payload using pKey + iv on client side.

PR #473, opened 2025-06-03
chore: add API for generating Hyperliquid address

jobotics:


src/config.ts
```diff
@@ -20,13 +21,15 @@ const configsByEnvironment: Record<NearIntentsEnv, EnvConfig> = {
     poaBridgeBaseURL: "https://bridge.chaindefuser.com",
     solverRelayBaseURL: "https://solver-relay-v2.chaindefuser.com",
     managerConsoleBaseURL: "https://api-mng-console.chaindefuser.com/api/",
+    hyperunitBaseURL: "https://api.hyperunit-testnet.xyz",
```
cawabunga-bytes: testnet by design or it is a mistake?

src/sdk/hyperunit/apis.ts
```diff
@@ -0,0 +1,36 @@
+import { config as globalConfig } from "../../config"
```
cawabunga-bytes: Let's not put this in sdk folder. This folder is specific to our protocol, it will export primitives to work with Intents.

cawabunga-bytes (changes requested):

dr-block (approved):

src/config.ts
jobotics: No, it's mistake, I'll update it

cawabunga-bytes (approved):

PR #474, opened 2025-06-03
feat: add Ton signatures support

cawabunga-bytes:


dr-block (approved):
commenting in case todo should be implemented within this PR

PR #476, opened 2025-06-04
feat: add hyperliquid network

jobotics:


dr-block (approved):

cawabunga-bytes (commented):
I am unsure about such changes. I suggest keeping BlockchainEnum name, since new name produces so much changes for no big reason.

PR #481, opened 2025-06-06
feat: display CEX checkbox for Polygon and BSC too

cawabunga-bytes:


jobotics (approved):

src/features/machines/withdrawFormReducer.ts
```diff
@@ -113,6 +114,9 @@ export const withdrawFormReducer = fromTransition(
           parsedRecipient: null,
           destinationMemo: "",
           parsedDestinationMemo: null,
+          cexFundsLooseConfirmation: isCexIncompatible(tokenOut)
```
dr-block: why not reuse 
isCexIncompatible(tokenOut) ? "not_confirmed" : "not_required",
as it appears in 3 places

PR #483, opened 2025-06-06
feat: add feature flag for Hyperliquid chain filtering

cawabunga-bytes:


jobotics (approved):

jobotics:
Is there a way we can fix the failing CI tests?

PR #492, opened 2025-06-11
fix: min amount validation considering fees

jobotics:


cawabunga-bytes (approved):

src/features/withdraw/components/WithdrawForm/components/RecipientSubForm/RecipientSubForm.tsx
```diff
@@ -113,6 +114,12 @@ export const RecipientSubForm = ({
 
   const onChangeNetwork = (network: SupportedChainName) => {
     setValue("blockchain", network)
+    actorRef.send({
+      type: "WITHDRAW_FORM.UPDATE_MIN_RECEIVED_AMOUNT",
+      params: {
+        minReceivedAmount: getMinWithdrawalHiperliquidAmount(network, tokenOut),
+      },
+    })
```
cawabunga-bytes, line 122: The chain is also may be automatically switched when token is changed. Please check out this case too.

jobotics: Yeah, I was thinking about this — it should only apply to the default token network. For SOL, ETH, and BTC, Hyperliquid isn’t the default network.

PR #498, opened 2025-06-16
feat: mixpanel events

dr-block:


src/features/gift/actors/giftTakerRootMachine.ts
```diff
@@ -85,6 +86,17 @@ export const giftTakerRootMachine = setup({
     setError: assign({
       error: (_, error: GiftTakerClaimingActorErrors) => error,
     }),
+    onClaimingDoneEntry: ({ context }) => {
```
jobotics: Nit: Consider renaming action to `emitGiftClaimed`.

src/features/machines/depositMachine.ts
```diff
@@ -86,6 +87,27 @@ export const depositMachine = setup({
         return params
       },
     }),
+    onSigningEntry: ({ context }) => {
```
jobotics: Nit: I would also rename this and the next action for better clarity.

src/features/machines/swapIntentMachine.ts
```diff
@@ -181,6 +182,31 @@ export const swapIntentMachine = setup({
         )
       }
     },
+    onSigningEntry: ({ context }) => {
```
jobotics: Same as above.

src/features/otcDesk/actors/otcMakerRootMachine.ts
```diff
@@ -181,6 +183,25 @@ export const otcMakerRootMachine = setup({
         iv: event.iv,
       })
     },
+    onSignedEntry: ({ context }) => {
```
jobotics: Consider renaming.

src/features/machines/withdrawUIMachine.ts
```diff
@@ -196,6 +197,23 @@ export const withdrawUIMachine = setup({
     clearPreparationOutput: assign({
       preparationOutput: null,
     }),
+    onPreparationEntry: ({ context }) => {
```
jobotics: Renaming, action only emits an event, and is unlikely to do more.

src/features/otcDesk/actors/otcMakerRootMachine.ts
```diff
@@ -413,6 +444,7 @@ export const otcMakerRootMachine = setup({
           },
         },
       },
+      entry: ["onSignedEntry"],
```
jobotics: You can move this event to the `signed` state and run it on entry, as this state indicates that the OTC trade has already been created. This way, you won't need `signerCredentials` in the context, since it will be accessible from `event.signerCredentials`.

dr-block: may be you meant action for onDone?

jobotics (approved):

PR #500, opened 2025-06-16
fix: set `add_public_key` gas to 5 TGas

Sliman4:
It absolutely doesn't need 300, all 300 does is require users to have 0.2 NEAR, which [not everyone has](https://t.me/intearchat/56827).


cawabunga-bytes (approved):
Thanks for the PR!

PR #505, opened 2025-06-19
chore: fetch TON token balances on active deposit

jobotics:
## Summary by CodeRabbit

- **New Features**
	- Added support for handling and displaying balances from the TON blockchain, including both native TON and Jetton tokens.
	- Introduced a new user wallet address field to enhance balance retrieval and deposit operations.

- **Bug Fixes**
	- Improved address validation and error handling for TON blockchain balance queries.

- **Chores**
	- Updated dependencies to include support for TON blockchain interactions.

### DIFF
--- a/src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
+++ b/src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
@@ -7,7 +7,7 @@ import { DepositUIMachineContext } from "./DepositUIMachineProvider"
 
 type DepositUIMachineFormSyncProviderProps = PropsWithChildren<{
   userAddress?: string
-  userWalletAddress?: string
+  userWalletAddress: string | null
   userChainType?: AuthMethod
 }>
 
@@ -59,7 +59,7 @@ export function DepositUIMachineFormSyncProvider({
   }, [watch, actorRef])
 
   useEffect(() => {
-    if (!userAddress || !userWalletAddress || userChainType == null) {
+    if (!userAddress || userChainType == null) {
       actorRef.send({
         type: "LOGOUT",
       })
--- a/src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
+++ b/src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
@@ -7,12 +7,14 @@ import { DepositUIMachineContext } from "./DepositUIMachineProvider"
 
 type DepositUIMachineFormSyncProviderProps = PropsWithChildren<{
   userAddress?: string
+  userWalletAddress?: string
   userChainType?: AuthMethod
 }>
 
 export function DepositUIMachineFormSyncProvider({
   children,
   userAddress,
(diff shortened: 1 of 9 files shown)


src/utils/validateAddress.ts
```diff
@@ -10,7 +10,8 @@ import { isLegitAccountId } from "./near"
 
 export function validateAddress(
   address: string,
-  blockchain: SupportedChainName
+  blockchain: SupportedChainName,
+  userWalletAddress?: string
```
cawabunga-bytes: this is not correct, this function just checks validity of the address, you should not modify it

src/features/machines/depositTokenBalanceMachine.ts
```diff
@@ -38,7 +41,7 @@ export const backgroundBalanceActor = fromPromise(
       nearBalance: null,
     }
 
-    if (!validateAddress(userAddress, blockchain)) {
+    if (!validateAddress(userAddress, blockchain, userWalletAddress)) {
```
cawabunga-bytes: It should be something like that `validateAddress(userWalletAddress ?? userAddress, blockchain)`

src/utils/validateAddress.ts
jobotics: Got you, I'll lift it up.

src/features/machines/depositTokenBalanceMachine.ts
jobotics: Yeap, got you point, thanks!

src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
```diff
@@ -57,17 +59,17 @@ export function DepositUIMachineFormSyncProvider({
   }, [watch, actorRef])
 
   useEffect(() => {
-    if (!userAddress || userChainType == null) {
+    if (!userAddress || !userWalletAddress || userChainType == null) {
```
cawabunga-bytes: This is not correct, passkeys won't have wallet address

src/features/machines/depositTokenBalanceMachine.ts
```diff
@@ -38,7 +41,7 @@ export const backgroundBalanceActor = fromPromise(
       nearBalance: null,
     }
 
-    if (!validateAddress(userAddress, blockchain)) {
+    if (!validateAddress(userWalletAddress ?? userAddress, blockchain)) {
```
cawabunga-bytes: btw this check doesn't make sense, it comes from the user's wallet and it's not from user input, so it will always be valid.

src/features/deposit/components/DepositUIMachineFormSyncProvider.tsx
jobotics: I did check it has `userAddress` which is something "p256:45BiCEwskZA2..." and `userWalletAddress` is the same. Otherwise we weren't login

cawabunga-bytes (approved):

PR #512, opened 2025-06-23
fix: deposit address missing user address concatenation

hyper54n3:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deposit processing by updating how deposit addresses are handled during transactions for increased reliability.


src/features/deposit/components/DepositUIMachineProvider.tsx
```diff
@@ -366,6 +366,8 @@ export function DepositUIMachineProvider({
 
                 assert(depositAddress != null, "Deposit address is required")
                 const chainId = getEVMChainId(chainName)
+                const finalDepositAddress =
```
jobotics: Nit: I think it should be `precompileDepositAddress`.

jobotics (approved):

PR #545, opened 2025-07-16
feat: support callback on swap token changes

jobotics:
## Summary by CodeRabbit

* **New Features**
  * Added support for an optional callback to notify external components when the input or output token changes in the Swap Widget.
  * The Swap Widget and its provider now accept an onTokenChange prop for real-time token change notifications.

### DIFF
--- a/src/features/swap/components/SwapUIMachineProvider.tsx
+++ b/src/features/swap/components/SwapUIMachineProvider.tsx
@@ -1,5 +1,6 @@
 import { createActorContext } from "@xstate/react"
 import type { PropsWithChildren, ReactElement, ReactNode } from "react"
+import { useRef } from "react"
 import { useFormContext } from "react-hook-form"
 import { assert } from "src/utils/assert"
 import { formatUnits } from "viem"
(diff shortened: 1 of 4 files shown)


src/features/swap/components/SwapUIMachineProvider.tsx
```diff
@@ -114,7 +121,31 @@ export function SwapUIMachineProvider({
...
+        onTokenChange={onTokenChange}
+        tokenIn={tokenIn}
+        tokenOut={tokenOut}
+      />
       {children}
     </SwapUIMachineContext.Provider>
   )
 }
+
+function TokenChangeNotifier({
```
cawabunga-bytes, line 134: This component is unnecessary, just use the hook directly, that's the whole point of them.

cawabunga-bytes (approved):

src/features/swap/components/SwapUIMachineProvider.tsx
jobotics: Yeah, I thought like this, but we can't use useTokenChangeNotifier outside the SwapUIMachineContext.Provider.

